PR #383, opened 2026-02-06
Support Aliases to refer to problems

mfornet:
With this feature we can refer to each problem by its aliases when doing `rbx on <NAME> <COMMAND>`. For <NAME> we can use either short name or any of the aliases.

I needed to bump the minimal python version required due to some syntax that was not available in python3.9.


mfornet:
It works now on python 3.9

PR #417, opened 2026-05-03
feat(generators): detect non-deterministic generators by re-running

rsalesc:
## Summary

- Run each generator-call testcase twice in parallel (via `asyncio.gather`) when `VerificationLevel >= VALIDATE` and compare output digests; mismatch raises `GenerationError`.
- The verification re-run streams stdout into a `DigestHolder` (no extra disk I/O); the original run continues writing to the on-disk testcase file.
- `verification` is plumbed through `generate_testcases` from `builder.build`. `BuildTestcaseVisitor` stays decoupled from rbx specifics — it only sees a `check_determinism: bool` flag.
- Catches generators that aren't seeded from `argv` (e.g. `time(NULL)`, `/dev/urandom`, unseeded `rand()`). Truly non-deterministic generators may still slip through if both runs coincide, but this raises the cost of being undetected.

Design: `docs/plans/2026-05-03-generator-determinism-check-design.md`
Plan: `docs/plans/2026-05-03-generator-determinism-check.md`

## Test plan

- [x] Non-deterministic generator at `VerificationLevel.VALIDATE` raises `GenerationError`
- [x] Deterministic generator at `VerificationLevel.VALIDATE` passes
- [x] Non-deterministic generator at `VerificationLevel.NONE` does **not** raise (gating works)
- [x] Pre-existing `tests/rbx/box/generators_test.py` still passes (25/25)
- [x] `ruff check` and `ruff format --check` clean

### DIFF
--- a/rbx/testdata/generators/gen-nondet.cpp
+++ b/rbx/testdata/generators/gen-nondet.cpp
@@ -1,0 +1,19 @@
+#include "testlib.h"
+#include <ctime>
+#include <unistd.h>
+
+using namespace std;
+
+int main(int argc, char *argv[]) {
+  registerGen(argc, argv, 1);
+  // Intentionally NOT seeded from argv: this generator is used in tests to
+  // verify the determinism check. It must produce different output on
+  // back-to-back runs.
+  static int nonce = 0;
+  unsigned long long v =
+      (unsigned long long)time(NULL) ^
+      (unsigned long long)getpid() ^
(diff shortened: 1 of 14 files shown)


rbx/testdata/generators/gen-nondet.cpp
rsalesc, line 1: can we make this a python script just to avoid the time we need to compile this file?

PR #467, opened 2026-05-18
feat(contest): show problem name in `rbx on` and `rbx each`

### LINKED ISSUE #466: `rbx on` and `rbx each` should show the name of the problems
At the moment, we show an UI with the letters / short_names of each problem. We should also show the name of the problem along with it.

rsalesc:
## Summary

Closes #466.

`rbx on` and `rbx each` previously labeled each problem in the command-runner UI with only its short name (`A`, `B`, `C`). This adds the problem name so entries read like `A. Two Sum`, which makes larger contests far easier to navigate.

## Changes

- Add `_problem_command_name(problem)` helper in `rbx/box/contest/main.py` that loads the problem package via the lenient `package.find_problem_package(...)` and returns `"<short_name>. <name>"`.
- Falls back to just the short name when the package can't be loaded (missing/broken `problem.rbx.yml`) or the name is empty, so the runner keeps working on a partially set-up contest.
- Used in both `each` and `on` (including `on`'s single-problem status line).
- Uses the problem **`name`** (not `titles`), per the issue intent.

## Testing

- New `TestProblemCommandName` unit tests cover: name present, package missing, empty name.
- `uv run pytest tests/rbx/box/contest/` — 85 passed.
- `uv run ruff check` / `ruff format` — clean.

### DIFF
--- a/rbx/box/contest/main.py
+++ b/rbx/box/contest/main.py
@@ -8,7 +8,7 @@ import syncer
 import typer
 
 from rbx import annotations, console, utils
-from rbx.box import cd, creation, presets, summary
+from rbx.box import cd, creation, package, presets, summary
 from rbx.box.contest import contest_package, contest_state, contest_utils, statements
 from rbx.box.contest.contest_package import (
     find_contest,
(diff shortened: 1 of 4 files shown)


rbx/box/contest/main.py
```diff
@@ -245,6 +245,19 @@ def remove(path_or_short_name: str):
     )
 
 
+def _problem_command_name(problem: ContestProblem) -> str:
```
rsalesc: Maybe move this to naming.py and assign a proper function name to convey what it's for.

rsalesc (commented):
PTAL

rbx/box/contest/main.py
rsalesc: Done in c7e7190 — moved to `rbx.box.naming.get_contest_problem_label(problem)` next to the other problem-naming helpers, and relocated the tests to `tests/rbx/box/test_naming.py`.

PR #477, opened 2026-05-23
feat(linters): built-in linter framework + side_effect C++ linter

### LINKED ISSUE #476: Add support for linters
We should add support for built-in linters in `rbx`. Notice this is slightly different from general linting support: this is to add support for rbx-made linters that are distributed as Python code.

We should have an interface that corresponds to linters. It returns warnings and errors, basically, and run given a certain code. Linters should be configured in `env.rbx.yml`, per language. We can have a `linters` field, that defaults to empty, where we can specify the linters that should be run. Every linter should have a string name (lowercase) that must be referenced there.

Linters can also be restricted to run only for certain classes of objects (generators, validators, etc). This can be defined both in the linter interface, and in the field in `env.rbx.yml`.

`linters: [side_effect]` or `linters: [{name: side_effect, applies_to: generators}]`

Notice when not specified, the linter applies to all assets (of course, also restricted by the restrictions given in the interface itself).

Warnings generated by linters should be added to the warning stack, errors generated by linters should result in a RbxException. Linters should always run only during the compilation phase.

---

We should also create the first linter, the `side_effect` linter. This linter should be a C++ linter that is implemented in Python through `tree-sitter-cpp`.

The idea of this linter is to identify the usage of side effect causing functions in function calls arguments. Function call arguments are essentially evaluated in an unspecified order. This might cause problems when the arguments themselves cause side-effects, because they can be evaluated in different orders than what was first expected.

We especifically saw this problem while writing a testlib-based generator that had `println(rnd.next(...), rnd.next(...))`, the numbers would get swapped under certain compilers.

I want this linter to focus specifically in functions with the name `rnd.next()`, coming from testlib, tgen and jngen, and also functions whose definitions are preceeded with a SIDE_EFFECT macro.

```cpp
// tgen.h

SIDE_EFFECT int fn_with_side_effect();
```

```cpp
// generator.h
#include "tgen.h"

// ...

some_function(fn_with_side_effect(), rnd.next()); // should warn
some_function(fn_with_side_effect(), 3); // should OK
some_function(rnd.next(), rnd.next()); // should warn
```

Notice we might have to expand into headers which might be non-trivial here. At first, we can leave the header expansion as a TODO in code and create an issue for it as there's a lot to discuss there.

rsalesc:
Closes #476.

## Summary

Adds a built-in **linter framework** to rbx plus the first linter, `side_effect`.

- **Framework** (`rbx/box/linters/`): a `Linter` ABC + name→instance registry. Linters are configured **per language** in `env.rbx.yml` via a new `linters` field, run during the **compilation phase** (inside `compile_item`), and produce messages routed by severity — `WARNING` → the existing warning stack, `ERROR` → `RbxException`.
- **Asset scoping**: a new `AssetKind` enum (`generator`, `validator`, `solution`, `checker`, `interactor`, `visualizer`). Kind is inferred from the `CodeItem` subclass where possible; validators are bare `CodeItem`s, so `compile_item` takes an explicit `kind` argument that `validators.py` passes. Linters can be restricted both in the interface (`applies_to`) and in config; the effective scope is their intersection.
- **Config forms** (both supported):
  ```yaml
  languages:
    - name: cpp
      linters:
        - side_effect                                   # shorthand
        - {name: side_effect, applies_to: [generators]} # restricted
  ```
- **`side_effect` linter** (`tree-sitter-cpp`): warns when a call passes **≥2 arguments** that each contain a known side-effecting call. C++ leaves argument evaluation order unspecified, which silently reordered `println(rnd.next(...), rnd.next(...))` under some compilers. v1 detects the hardcoded `rnd.next` family (testlib/tgen/jngen).

## Scope decisions (from the design doc)
- `side_effect` emits **warnings** by default; the interface still supports errors for future linters.
- v1 is **hardcoded `rnd.next` only**. The `SIDE_EFFECT` macro + header expansion described in the issue are **deferred** (TODO in `side_effect.py`); a follow-up issue should be opened.

Design and plan: `docs/plans/2026-05-23-linters-design.md`, `docs/plans/2026-05-23-linters.md`.

## Behavior on the issue's examples
```cpp
some_function(rnd.next(), rnd.next());            // WARN  (2 side-effect args)
some_function(rnd.next(), 3);                     // OK    (1 side-effect arg)
some_function(fn_with_side_effect(), rnd.next()); // OK in v1 (macro not yet detected)
```
`cout << rnd.next() << rnd.next()` is not flagged (`<<` operands are sequenced). Nested calls are covered.

## Dependencies
Adds `tree-sitter` and `tree-sitter-cpp` (resolved to 0.25.2 / 0.23.4).

## Testing
- 23 new tests in `tests/rbx/box/linters/` — all pass.
- `ruff format`/`ruff check` clean.
- The default env ships **no** `linters` config, so `run_linters` returns immediately on existing code paths (verified no-op).
- Full suite: pre-existing C++ compile/sandbox/checker/validator/docker failures on the local machine are unrelated to this change; no new failures introduced.

## Follow-up
Open an issue for `SIDE_EFFECT` macro detection + header expansion (referenced by the TODO in `side_effect.py`).

### DIFF
--- a/rbx/box/linters/cpp/side_effect.py
+++ b/rbx/box/linters/cpp/side_effect.py
@@ -1,0 +1,92 @@
+from typing import List, Set, Tuple
+
+import tree_sitter_cpp
+from tree_sitter import Language, Node, Parser
+
+from rbx.box.linters import registry
+from rbx.box.linters.linter import Linter, LinterMessage, LinterSeverity
+from rbx.box.schema import CodeItem
+
+# Known side-effecting calls, as (object, method) pairs matching `obj.method(...)`.
+# Seeded with the testlib/tgen/jngen `rnd.next` family. Extend as needed.
+# TODO: SIDE_EFFECT macro detection + header expansion (follow-up to #476)
+# Detect SIDE_EFFECT-macro-annotated functions and expand headers so
+# side-effecting functions declared in tgen.h/jngen.h/testlib.h are seen.
+SIDE_EFFECT_CALLS: Set[Tuple[str, str]] = {
+    ('rnd', 'next'),
+}
+
+_LANGUAGE = Language(tree_sitter_cpp.language())
+
+
+def _parser() -> Parser:
+    return Parser(_LANGUAGE)
+
+
+def _text(node: Node) -> str:
+    return node.text.decode('utf8')
+
+
+def _is_side_effect_call(node: Node) -> bool:
+    """True if `node` is a call_expression to a known side-effecting function."""
+    if node.type != 'call_expression':
+        return False
+    fn = node.child_by_field_name('function')
+    if fn is None or fn.type != 'field_expression':
+        return False
+    obj = fn.child_by_field_name('argument')
+    field = fn.child_by_field_name('field')
+    if obj is None or field is None:
+        return False
+    return (_text(obj), _text(field)) in SIDE_EFFECT_CALLS
+
+
+def _contains_side_effect_call(node: Node) -> bool:
+    if _is_side_effect_call(node):
+        return True
+    return any(_contains_side_effect_call(child) for child in node.named_children)
+
+
+def _argument_nodes(call: Node) -> List[Node]:
+    args = call.child_by_field_name('arguments')
+    if args is None:
+        return []
+    return list(args.named_children)
+
+
+class SideEffectLinter(Linter):
+    name = 'side_effect'
+    languages = {'cpp'}
+    applies_to = set()  # all kinds; restrictable via env config
+
+    def lint(self, code: CodeItem, source: str) -> List[LinterMessage]:
+        tree = _parser().parse(bytes(source, 'utf8'))
+        messages: List[LinterMessage] = []
+        self._visit(tree.root_node, messages)
+        return messages
+
+    def _visit(self, node: Node, messages: List[LinterMessage]) -> None:
+        if node.type == 'call_expression':
+            side_effect_args = sum(
+                1 for arg in _argument_nodes(node) if _contains_side_effect_call(arg)
+            )
+            if side_effect_args >= 2:
+                row, col = node.start_point
+                messages.append(
+                    LinterMessage(
+                        severity=LinterSeverity.WARNING,
+                        message=(
+                            'Call passes multiple side-effecting arguments '
+                            '(e.g. rnd.next()); C++ leaves argument evaluation '
+                            'order unspecified, so results may differ across '
+                            'compilers. Compute each value into a variable first.'
+                        ),
+                        line=row + 1,
+                        col=col + 1,
+                    )
+                )
+        for child in node.named_children:
+            self._visit(child, messages)
+
+
+registry.register(SideEffectLinter)
--- a/rbx/box/linters/linter.py
+++ b/rbx/box/linters/linter.py
@@ -21,10 +21,10 @@ class LinterMessage(BaseModel):
 
 
 class Linter(abc.ABC):
-    # Lowercase identifier referenced in env.rbx.yml.
+    # Lowercase identifier referenced in env.rbx.yml. A linter runs for whatever
+    # language entry it is configured under in env.rbx.yml; it is the user's
+    # responsibility to add it under languages where it makes sense.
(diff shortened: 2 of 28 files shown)


rbx/box/linters/linter.py
```diff
@@ -0,0 +1,34 @@
...
+    message: str
+    line: Optional[int] = None  # 1-based
+    col: Optional[int] = None  # 1-based
+
+
+class Linter(abc.ABC):
+    # Lowercase identifier referenced in env.rbx.yml.
+    name: ClassVar[str]
+    # Env language names this linter supports (e.g. {'cpp'}).
+    languages: ClassVar[Set[str]]
```
rsalesc: This is not necessary, linters should be applied to whatever languages they're added to in the `env.rbx.yml`, and it's the responsibility of the user to put it wherever it makes sense.

rbx/box/linters/cpp/side_effect.py
```diff
@@ -0,0 +1,92 @@
...
+    args = call.child_by_field_name('arguments')
+    if args is None:
+        return []
+    return list(args.named_children)
+
+
+class SideEffectLinter(Linter):
+    name = 'side_effect'
+    languages = {'cpp'}
+    applies_to = set()  # all kinds; restrictable via env config
```
rsalesc: let's apply this only to generators

rbx/box/linters/cpp/side_effect.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _argument_nodes(call: Node) -> List[Node]:
+    args = call.child_by_field_name('arguments')
+    if args is None:
+        return []
+    return list(args.named_children)
+
+
+class SideEffectLinter(Linter):
+    name = 'side_effect'
```
rsalesc: Let's fully rename this linter to be called "testlib" linter. One of the things it should check for is for the rnd.next calls, and in the future we'll add the others.

This name makes more sense.

rsalesc (commented):
PTAL

rbx/box/linters/linter.py
rsalesc: Removed the `languages` ClassVar and its guard in `runner.py` (74f9df5). Linters now run for whatever language entry they are configured under. While here I also fixed `_in_scope`: with no non-empty interface scope before, a disjoint interface∩config intersection was wrongly treated as "applies to all" — it now correctly means "never applies".

rbx/box/linters/cpp/side_effect.py
rsalesc: Done — interface `applies_to = {AssetKind.GENERATOR}` (b... in the rename commit). Added a test that a disjoint config scope (e.g. `applies_to: [solutions]`) does not run it on a generator.

rbx/box/linters/cpp/side_effect.py
rsalesc: Renamed to the `testlib` linter: `TestlibLinter` / `name = "testlib"`, file `cpp/side_effect.py` → `cpp/testlib.py`, tests and docs updated. The `rnd.next` argument-ordering check is now framed as its first check, with room for more testlib-specific checks later.

rsalesc:
Addressed all three review comments (replies in-thread):
- Removed the per-linter `languages` restriction; linters run for whatever language they're configured under. Also fixed a latent `_in_scope` bug where disjoint interface/config `applies_to` scopes were treated as "applies to all".
- Restricted the linter to generators at the interface level.
- Renamed `side_effect` → `testlib` linter (`TestlibLinter`, `cpp/testlib.py`); the `rnd.next` check is now its first check.

On the red `tests` check: these failures are pre-existing/environmental, not from this PR. The `tests` workflow runs only on PRs/tags and is red on every recent run — including the `0.36.0` tag push on `main` (28 failures) and merged PRs (#475: 12). The failing sets vary run-to-run (flaky sandbox/compile under `-n auto`) and are all testlib/sandbox C++ compile errors; this PR's source diff doesn't touch testlib, and the linter is a no-op when no `linters` are configured (the default env configures none). All 24 linter tests pass locally and `ruff` is clean.

PR #496, opened 2026-06-01
fix(boca): emit exact fractional time limits, deprecate rounding (#494)

### LINKED ISSUE #494: Deprecate BOCA rounding + approximation
Although BOCA/safeexec supports specifying fractional time limits, for some historical reason, packages are usually built with integer time limits and multiple runs. For instance, if a problem TL is 0.1s, we can simply run it 10 times with the BOCA TL of 1s.

This works just fine when TLs are round, such as with 0.2s, 0.5s, 1.5s, as in these cases we can simply specify a small, reasonable number of runs without losing precision over the original time limit.

Also, it's fine to specify in the infosheet that the TL is the actual fractional value, as in practice, running a program N times with N*TL is equivalent.

Things get fuzzy when the estimated TL is not so pretty, such as 0.3s, 0.7s, 1.2s or anything like this. We can still run the program 10 times, but there are heuristics in place that consider this too much, and end up falling back to a small integer time that would allow for an error of at most X% on the original TL.

This heuristic is inherited from box for feature parity.

This caused problems because rbx estimates TLs based on a (customizable) formula that defaults to allowing for TLs with 0.1s resolution (i.e. multiples of 0.1s). TLs such as 1.2, 3.9s would then get approximated while allowing for errors, which would in practice change the TL.

Why this wasn't an issue in PdA and Maratona Cearense?

I actually have no idea whether this happened in Cearense. In PdA, this didn't happen for one very specific reason: we changed the time limit formula, and thus the time resolution, to 0.5s. Thus, there was no TL that was not a multiple of 0.5s. This is a very happy case, where almost all problems can evaluated with a precise TL with either 1 or 2 runs.

How to solve?

Get rid of the time approximation entirely and introduce a new timing mechanism. Allow for fractional time limits in all packages. If the setter wants, they can establish a minimum running time for their problems.

For instance, one can set the minimum running time to 1s, and problems with 0.3s will be run 4 times with a 1.2s TL. This is fine and equivalent, but we should never approximate/round/allow for TL errors. This is a box feature that should NOT be ported to rbx.

rsalesc:
## Summary

Fixes #494. The BOCA packager previously rounded time limits to an **integer** budget within a 20% `maximumTimeError` tolerance, then ran the solution multiple times to approximate the real TL. Because BOCA's effective per-run TL is `budget / repetitions`, rounding the budget silently shifted the effective TL whenever the product wasn't near-integer:

| TL | old effective TL | drift |
|----|------------------|-------|
| 1.2s | 1.0s | **−16.7%** (intended-AC solutions wrongly TLE) |
| 0.3s | 0.333s | **+11.1%** (intended-TLE solutions wrongly pass) |
| 1.9s | 2.0s | +5.3% |

This only stayed hidden on past contests (e.g. PdA) because they used a 0.5s TL-formula resolution, which always lands on near-integer products. rbx's default 0.1s resolution routinely produces "ugly" TLs like 1.2s/3.9s.

BOCA/safeexec supports fractional time budgets (rbx already emits them for interactive tasks), so the rounding is unnecessary and is removed.

## What changed

- **Exact fractional budgets, no rounding.** New `_fmt_seconds(ms)` formats integer milliseconds as exact fractional seconds via integer arithmetic (`f'{ms//1000}.{ms%1000:03d}'`) — no float error. Used by both BATCH and COMMUNICATION paths (the latter previously used a lossy `:.2f`).
- **Optional `minRunningTime` (ms) floor** on `BocaExtension`. When set, the solution runs `ceil(minRunningTime / timeLimit)` times so the accumulated budget reaches the floor (e.g. min=1000ms, TL=300ms → 4 runs, 1.200s budget), while the effective per-run TL stays **exactly** the real TL. Capped at 10 runs with a warning; unset → a single run.
- **`maximumTimeError` deprecated and ignored** (kept as a field for back-compat — no schema break).
- Extracted pure helpers `_fmt_seconds` and `_compute_reps` for direct unit testing; removed dead code (`test_time`, `_MAX_REP_TIME`, `_MAX_REP_ERROR`, `from math import fabs`).

MOJ is unaffected — it already emits fractional TLs directly.

## Testing

- New unit tests for `_fmt_seconds` (exactness across edge cases), `_compute_reps` (ceil, cap+flag, no-minimum, non-positive-TL guard), and `_get_limits` (asserts all four echo lines; the 1200ms case fails against the old rounding code — regression guard).
- `tests/rbx/box/packaging/boca` + `boca_next`: **145 passed**.
- Full `tests/rbx/box/packaging` slice: 148 passed, 1 xfailed (the 3 docker-compose e2e errors are environmental/pre-existing, unrelated to this change).

Design + implementation notes: `docs/plans/2026-06-01-boca-exact-fractional-tl-design.md`, `docs/plans/2026-06-01-boca-exact-fractional-tl.md`.

### DIFF
--- a/rbx/box/packaging/boca/extension.py
+++ b/rbx/box/packaging/boca/extension.py
@@ -4,13 +4,21 @@ from pydantic import BaseModel, Field
 
 BocaLanguage = typing.Literal['c', 'cpp', 'cc', 'kt', 'java', 'py2', 'py3']
 
-_MAX_REP_ERROR = 0.2  # 20% error allowed in time limit when adding reps
-
 
 class BocaExtension(BaseModel):
     languages: typing.List[BocaLanguage] = []
     flags: typing.Dict[BocaLanguage, str] = {}
-    maximumTimeError: float = _MAX_REP_ERROR
+    # Optional floor (in milliseconds) on the TOTAL BOCA time budget. When set, the
+    # solution is run ceil(minRunningTime / timeLimit) times so the accumulated budget
+    # reaches the floor, amortizing fixed startup/JIT overhead and measurement noise on
+    # small TLs. The effective per-run TL always stays exactly equal to the real TL.
+    minRunningTime: typing.Optional[int] = Field(default=None, gt=0)
+    # Deprecated (issue #494): BOCA/safeexec supports fractional time budgets, so rbx no
+    # longer rounds TLs. This field is ignored; use `minRunningTime` instead.
+    maximumTimeError: typing.Optional[float] = Field(
+        default=None,
+        deprecated='Ignored since #494; rbx emits exact fractional TLs. Use minRunningTime.',
+    )
     preferContestLetter: bool = False
     usePypy: bool = False
 
--- a/rbx/box/packaging/boca/extension.py
+++ b/rbx/box/packaging/boca/extension.py
@@ -10,8 +10,7 @@ class BocaExtension(BaseModel):
     flags: typing.Dict[BocaLanguage, str] = {}
     # Optional floor (in milliseconds) on the TOTAL BOCA time budget. When set, the
     # solution is run ceil(minRunningTime / timeLimit) times so the accumulated budget
-    # reaches the floor, amortizing fixed startup/JIT overhead and measurement noise on
-    # small TLs. The effective per-run TL always stays exactly equal to the real TL.
+    # reaches this floor, while the effective per-run TL stays exactly equal to the real TL.
     minRunningTime: typing.Optional[int] = Field(default=None, gt=0)
     # Deprecated (issue #494): BOCA/safeexec supports fractional time budgets, so rbx no
     # longer rounds TLs. This field is ignored; use `minRunningTime` instead.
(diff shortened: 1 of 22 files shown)


rbx/box/packaging/boca/extension.py
```diff
@@ -4,13 +4,21 @@
...
-_MAX_REP_ERROR = 0.2  # 20% error allowed in time limit when adding reps
-
 
 class BocaExtension(BaseModel):
     languages: typing.List[BocaLanguage] = []
     flags: typing.Dict[BocaLanguage, str] = {}
-    maximumTimeError: float = _MAX_REP_ERROR
+    # Optional floor (in milliseconds) on the TOTAL BOCA time budget. When set, the
+    # solution is run ceil(minRunningTime / timeLimit) times so the accumulated budget
+    # reaches the floor, amortizing fixed startup/JIT overhead and measurement noise on
```
rsalesc: This has nothing to do with this

rsalesc: Good catch — removed the unfounded rationale (startup/JIT/noise wasn't the actual purpose). The comment now only describes the mechanic: it's an optional floor on the total budget that drives the repetition count. Fixed in 7a33453 (and the matching lines in the design/plan docs).

Issue #493, opened 2026-06-01
BOCA language name mismatch during time estimation


rsalesc:
What went wrong?

In older versions of BOCA, cpp is actually called cc. Both are actually supported by rbx, so the package is built for both language variants, with very similar flags.

The problem is that the TL for cpp is taken from the actual cpp rbx language, but cc is not found in .limits/boca.yml, and then the TL is taken from the fallback value in .limits/boca.yml, which is potentially different.

How to solve?

This can be somewhat more easily solved with the introduction of https://github.com/rsalesc/rbx/issues/453

We should simply assume cc == cpp and take time limits from the actual underlying rbx entry.

PR #498, opened 2026-06-01
feat: configurable per-language wall time limits (#490)

### LINKED ISSUE #490: Wall time limit is too tight for slow languages
Add ability to configure wall time limits in the rbx-level for each language specifically, and set sensible default values for them in the default presets.

We should support some kinda of `ax+b` formula, where `x` is the default TL.

Something like `wallTimeMultiplier` and `wallTimeIncrement`.

rsalesc:
## Summary

Closes #490. Wall time limits were too tight for slow languages (Java/Kotlin/Python), where JVM/interpreter startup eats into the wall-clock budget even when CPU time is fine — and there was no per-language knob. The wall time was hardcoded as `cpu_TL × 2` in local judging and `ttime = $rtime + 30` in the BOCA run scripts (the `+30` being an ad-hoc Maratona Mineira hack, since superseded by exact fractional limits in #494).

This adds a configurable `wall = a·x + b` formula, shared by **rbx local judging** and **BOCA packaging**, where:
- `a` = `wallTimeMultiplier` (≥ 1.0, default `2.0`)
- `b` = `wallTimeIncrement` in ms (≥ 0, default `0`)
- `x` = the effective per-language CPU time limit (after per-language modifiers and double-TL expansion)

Env-wide defaults live in the `timing` block of `env.rbx.yml`; each language can override either coefficient via its own `timing` field (unspecified coefficients fall back to the env default).

### Changes
- **`environment.py`**: `wallTimeMultiplier`/`wallTimeIncrement` on `TimingConfig` and a new `LanguageTimingConfig`; `EnvironmentLanguage.timing`; shared `resolve_walltime_coeffs` / `apply_walltime_formula` / `compute_walltime`.
- **`tasks.py`**: `_get_execution_config` applies `compute_walltime` in the soft-timeout branch (replaces hardcoded `× 2`); all call sites thread the language through.
- **BOCA** (`packager.py` + 14 `run/`/`interactive/` templates): `_replace_walltime` maps the emitted BOCA language to its rbx language (reusing #493's helper) and substitutes the coefficients; the shell multiplies the exact fractional `$time` directly for accuracy. The degenerate `rtime ≤ 0` branch keeps a wall floor.
- **Default preset**: `wallTimeMultiplier: 2.0`, `wallTimeIncrement: 1000`, with `py: +2000`, `java`/`kt: +3000`.
- **Docs**: "Wall time limits" section in the Environment reference.

Defaults are chosen so rbx local judging is **unchanged** (`2.0×, +0` at the schema level reproduces today's `× 2`); the preset adds modest slack. The BOCA cushion changes from a flat `+30 s` to `2·x + 1 s` (plus per-language slack) — an intended replacement of the old hack.

Design + implementation plan: `docs/plans/2026-06-01-configurable-walltime-design.md`, `docs/plans/2026-06-01-configurable-walltime.md`.

## Test Plan
- [x] `tests/rbx/box/walltime_test.py` — schema constraints, field-by-field override precedence, formula truncation, end-to-end resolution against the shipped preset (py=4000, java/kt=5000), `_get_execution_config` wiring incl. double-TL.
- [x] `tests/rbx/box/packaging/boca/test_timing.py` — substitution against real templates, cc/cpp mapping equivalence, per-language override, degenerate else-branch floor.
- [x] Broader sweep: 223 passed across `tests/rbx/box/packaging`, `tasks_test`, `limits_info_test`, `test_timing` (3 BOCA e2e errors are Docker-infra only, pre-existing).
- [ ] Manual: package a problem with Java/Python solutions for BOCA and confirm emitted `run/*` wall times.

### DIFF
--- a/rbx/resources/packagers/boca/interactive/c
+++ b/rbx/resources/packagers/boca/interactive/c
@@ -91,12 +91,12 @@ if [ "$4" != "" ]; then
   fi
 fi
 rtime=$(awk "BEGIN {print int($time+0.9999999)}")
-if [ "$rtime" -gt "0" ]; then
-  ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} * $nruns + 0.9999999)}")
-else
+# A non-positive (rounded) time limit means "no real limit"; force a 1s CPU
+# budget so the wall formula below still yields a sane positive wall time.
+if [ "$rtime" -le "0" ]; then
   time=1
-  ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} * $nruns + 0.9999999)}")
 fi
+ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} * $nruns + 0.9999999)}")
 
 maxm=512000
 if [ "$5" != "" ]; then
--- a/rbx/resources/packagers/boca/interactive/c
+++ b/rbx/resources/packagers/boca/interactive/c
@@ -89,7 +89,7 @@ if [ "$rtime" -gt "0" ]; then
   ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} + 0.9999999)}")
 else
   time=1
-  ttime=$(awk "BEGIN {print int({{rbxWallIncrement}}+0.9999999)}")
+  ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} + 0.9999999)}")
 fi
 
 nruns=1
--- a/rbx/resources/packagers/boca/interactive/c
+++ b/rbx/resources/packagers/boca/interactive/c
@@ -86,10 +86,10 @@ fi
 time=$3
 rtime=$(awk "BEGIN {print int($time+0.9999999)}")
 if [ "$rtime" -gt "0" ]; then
-  let "ttime = $rtime + 30"
+  ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} + 0.9999999)}")
 else
   time=1
-  ttime=30
+  ttime=$(awk "BEGIN {print int({{rbxWallIncrement}}+0.9999999)}")
 fi
 
 nruns=1
--- a/rbx/resources/packagers/boca/interactive/c
+++ b/rbx/resources/packagers/boca/interactive/c
@@ -84,20 +84,20 @@ if [ ! -x "$sf" ]; then
 fi
 
 time=$3
-rtime=$(awk "BEGIN {print int($time+0.9999999)}")
-if [ "$rtime" -gt "0" ]; then
-  ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} + 0.9999999)}")
-else
-  time=1
-  ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} + 0.9999999)}")
-fi
-
 nruns=1
 if [ "$4" != "" ]; then
   if [ "$4" -gt "1" ]; then
     echo "WARNING: nruns is set to $4, but it will be ignored because it's not supported by interactive problems"
   fi
 fi
+rtime=$(awk "BEGIN {print int($time+0.9999999)}")
+if [ "$rtime" -gt "0" ]; then
+  ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} * $nruns + 0.9999999)}")
+else
+  time=1
+  ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} * $nruns + 0.9999999)}")
+fi
+
 maxm=512000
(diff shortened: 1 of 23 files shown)


rbx/resources/packagers/boca/interactive/cc
```diff
@@ -86,10 +86,10 @@ fi
 time=$3
 rtime=$(awk "BEGIN {print int($time+0.9999999)}")
 if [ "$rtime" -gt "0" ]; then
-  let "ttime = $rtime + 30"
+  ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} + 0.9999999)}")
```
rsalesc: Could you verify whether ttime really needs to be an integer, of it's ok to be passed as a floating point downstream?

rsalesc: Let's multiply the increment by the number of runs since it directly affects/amplifies the overhead introduced by the language in each run.

rsalesc: Verified — `ttime` must be an integer. The compiled `safeexec` parses the wall-time flag `-T` with `atoi` (`safeexec.c:347`: `case 'T': real_timeout = atoi(optarg);`), and `real_timeout` is an integer second counter used by the watchdog (`safeexec.c:245`). A fractional value like `3.7` would be truncated by `atoi` to `3` — i.e. it would silently *floor* the wall limit.

(Note the asymmetry: the CPU flag `-t` *does* accept fractional seconds — `safeexec.c:344-345` uses `atof`/`cpu_timeoutdouble` — which is why we pass the exact fractional `$time` for `-t`.)

So we keep the ceiling `int($time * mult + incr + 0.9999999)` to round wall time *up* to whole seconds, never under-allocating. No change needed here.

rsalesc: Done in 951c53e. Both `ttime` branches now use `{{rbxWallIncrement}} * $nruns`, and the `nruns` block was moved above the `ttime` computation in all 14 run/interactive templates so `$nruns` is in scope.

Since `$time` is the *aggregated* CPU budget (`per-run TL × nruns`, from `_get_limits`), the multiplier term `$time * mult` already scales with runs; only the additive per-run startup overhead `b` needed `× $nruns`. The wall is now `mult·tl·nruns + b·nruns = nruns·(mult·tl + b)` — exactly `nruns ×` the rbx per-run wall, so both paths agree per run. (Interactive forces `nruns=1`, so it's a no-op there.)

rsalesc:
### Re: do we need to double the WTL for interactive problems?

Investigated the local communication path (`stupid_sandbox.execute_communication`, `program.py` wall kill, `checkers.check_communication`). Findings:

- The solution and interactor are killed **independently**, each on its own wall limit — there's no shared clock.
- The verdict comes from the **solution's** run log (wall-TLE → `IDLENESS_LIMIT_EXCEEDED`). The interactor's wall limit should **never** be what fires: if the interactor times out first, the sandbox's asymmetric group-kill (`stupid_sandbox.py:337-341`) tears down a still-legal solution and surfaces a bogus interactor failure.
- So the interactor only needs to **outlive the solution** — `interactor_wall > solution_wall` plus a small margin. It does **not** need (and shouldn't get) double the budget.

The old code summed the two configs: `interactor_wall = (a·cpu + b) + (a·cpu + b)`, which double-counted both `a·cpu` and `b` **and** applied the solution's per-language coefficients to the (usually C++) interactor. That over-allocates badly with large `a`/TL (a 5 s TL gave a 20 s interactor wall) and just delays detection of a hung interactor.

This is exactly the model BOCA already uses: `ittime = ttime + 1` (solution wall + 1 s), see `interactor_run.sh:13`.

**Fixed in 43b83c0**: the interactor now gets `solution_wall + 1000 ms` (a fixed `_INTERACTOR_WALL_MARGIN_MS`), built without the solution's language. The solution's own wall is unchanged, so no normal-solution verdicts change — it only stops over-allocating the interactor and aligns local judging with BOCA. So: **no, we don't double it.**

rbx/resources/packagers/boca/interactive/c
```diff
@@ -84,20 +84,20 @@ if [ ! -x "$sf" ]; then
...
-  ttime=30
-fi
-
 nruns=1
 if [ "$4" != "" ]; then
   if [ "$4" -gt "1" ]; then
     echo "WARNING: nruns is set to $4, but it will be ignored because it's not supported by interactive problems"
   fi
 fi
+rtime=$(awk "BEGIN {print int($time+0.9999999)}")
```
rsalesc, line 93: There seems to be some code duplicaton here, can this be solved?

rsalesc: Fixed in 13a2cf3. After the earlier else-branch fix both branches computed the identical `ttime` awk, so the only branch-specific action was forcing `time=1`. Deduplicated to compute `ttime` once:

```bash
rtime=$(awk "BEGIN {print int($time+0.9999999)}")
# A non-positive (rounded) time limit means "no real limit"; force a 1s CPU
# budget so the wall formula below still yields a sane positive wall time.
if [ "$rtime" -le "0" ]; then
  time=1
fi
ttime=$(awk "BEGIN {print int($time * {{rbxWallMultiplier}} + {{rbxWallIncrement}} * $nruns + 0.9999999)}")
```

Exactly equivalent (rtime>0 → `formula($time)`; rtime≤0 → `time=1` then `formula(1)`), applied to all 14 run/interactive templates (`rbxWallMultiplier` now appears once per file instead of twice). Tests updated accordingly.

PR #499, opened 2026-06-02
feat(timing): language groups for time-limit estimation (#497)

### LINKED ISSUE #497: `rbx time` cause problems for unrepresented languages
## What went wrong?

Languages that had no solutions for a problem will usually default to the non-language-specific TL in `.limits/boca.yml`.

Let's suppose we have a problem with solutions in C++, Python and Kotlin. If we run rbx time for those and set specific times for these ALL THESE THREE languages, languages such as C and Java would fall back to a non-reasonable time limit, which could cause unexpected problems in the contest, such as these languages having tight time limits.

It's hard to notice this happens because when you set specific time limits for all languages, you end up with a totally different default/fallback.

## How to solve?

First, we should allow users to bucketize languages when estimating TLs (such as C/C++ and Java/Kotlin). This can be something configurable in the interactive CLI command itself, similar to how we ask what languages should have specific time limits, we should be able to group them together as well, in a fancy way.

We can also provide a default grouping for them in `env.rbx.yml` (notice this is judge agnostic, probably should live in the `timing` message and refer to the actual `rbx` languages for grouping).

Also, we should provide a very detailed table, or any other representation, that makes it VERY clear what is the TL for each language/bucket after the command runs, and also show the default time limit.

Additionally (can be part of a different task or issue), we should also provide this very detailed table when building packages that consume from those limits file.

For instance, `.limits/boca.yml` is consumed by the BOCA packager. At the moment of packaging, as a refresher, we should show this very same table/visual representation of the TLs so it is very clear what is being used as TLs in there.

rsalesc:
## Summary

Fixes #497 — languages without a solution silently defaulted to a (potentially tight) base time limit, invisibly, because setting per-language limits masked the fallback.

This introduces **language groups** for time-limit estimation (Approach A: groups are an estimation-time abstraction that compiles down to the existing per-language `modifiers`, so no consumer's resolution path changes):

- **`env.rbx.yml` `timing.groups`** — anonymous, disjoint groups of related languages (e.g. `[c, cpp]`, `[java, kotlin]`); unlisted languages are implicit singletons. Each group may declare `whenEmpty: {relativeTo: <language>, multiplier: >0}`, used only when the group has no solutions (multiplies the referenced group's TL, transitively; or the base estimate if `relativeTo` is omitted).
- **`rbx time`** pools accepted-solution timings per group, estimates a TL per non-empty group, resolves empty groups via `whenEmpty` (or falls back to the base TL with a **loud DEFAULTED warning**), and expands everything into `modifiers[lang].time`. An interactive group-number prompt (prepopulated from `env.rbx.yml`; `--auto` skips it) lets you repartition per run.
- **Presentation** — a shared per-group table (Languages / Solutions / Time Limit / Source) is shown after `rbx time` and at the end of `rbx package boca` (rendered from the resolved profile so inherit/custom profiles show real limits, not `0 ms`). DEFAULTED rows are highlighted.
- **Metadata** — an optional, presentation-only `groups` field on `LimitsProfile` records the grouping so the table is reconstructable from a saved profile; old `.limits/*.yml` files load unchanged.

Pure grouping logic lives in `rbx/box/timing_groups.py` (partition / resolve / validate / repartition), fully unit-tested without the sandbox.

Design & plan: `docs/plans/2026-06-01-language-groups-timing-design.md`, `docs/plans/2026-06-01-language-groups-timing.md`.

## Test Plan

- [x] New unit tests: `test_timing_groups.py`, `test_environment_groups.py`, `test_limits_profile_groups.py`, `test_limits_table.py`, `test_timing_estimation.py`, `packaging/test_boca_limits_table.py`
- [x] Updated existing `test_timing.py` for the new estimation flow
- [x] All directly-relevant suites pass (108 tests); `ruff check`/`format` clean repo-wide; `mkdocs build` succeeds
- [x] Disjoint-group validation, `whenEmpty` ref/self/cycle rejection, `multiplier > 0`, transitive multiplier chains, DEFAULTED fallback, and degraded (no-metadata) table rendering all covered
- [ ] Manual: run `rbx time -p boca` then `rbx package boca` on a multi-language problem with `timing.groups` configured and confirm the per-group table + DEFAULTED highlighting

> Note: broad `tests/rbx/box` sweep shows pre-existing failures confined to checker/validator/generator/sandbox suites (require a working C++ toolchain/sandbox); none are in the modules changed here.

### DIFF
--- a/rbx/box/timing.py
+++ b/rbx/box/timing.py
@@ -129,13 +129,11 @@ async def _prompt_repartition(
 def relevant_languages_for_estimation(
     env_languages: List[str],
     timing_languages: List[str],
-    env_groups: List[environment.LanguageGroup],
 ) -> List[str]:
     """Languages that participate in the partition during estimation: every
     environment language (so unrepresented ones land in the picker and the
     leftover pool / DEFAULTED warning), followed by any timing language not
     declared in the environment. Ordered by the environment's language order."""
-    del env_groups  # no longer needed; all env languages are in scope
     ordered = list(env_languages)
     for lang in timing_languages:
         if lang not in ordered:
@@ -201,7 +199,6 @@ async def estimate_time_limit(
     all_languages = relevant_languages_for_estimation(
         env_languages=[lang.name for lang in env.languages],
         timing_languages=list(timing_per_solution_per_language.keys()),
-        env_groups=env_groups,
     )
 
     repartition = None
--- a/rbx/box/timing.py
+++ b/rbx/box/timing.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 import questionary
 import rich
@@ -8,7 +8,7 @@ from ordered_set import OrderedSet
 from pydantic import BaseModel, Field
 
 from rbx import console, utils
-from rbx.box import environment, limits_info, package, safeeval, schema
+from rbx.box import environment, limits_info, package, safeeval, schema, timing_groups
 from rbx.box.code import find_language_name
 from rbx.box.environment import VerificationLevel
 from rbx.box.formatting import href
@@ -26,6 +26,7 @@ class TimingProfile(BaseModel):
     timeLimit: int
     formula: Optional[str] = None
     timeLimitPerLanguage: Dict[str, int] = Field(default_factory=dict)
+    groups: Optional[List[schema.TimingGroupReport]] = None
 
     def to_limits(self):
         return schema.LimitsProfile(
@@ -35,6 +36,7 @@ class TimingProfile(BaseModel):
                 lang: schema.LimitModifiers(time=tl)
                 for lang, tl in self.timeLimitPerLanguage.items()
             },
+            groups=self.groups,
         )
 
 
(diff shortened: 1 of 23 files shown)


rbx/box/timing.py
```diff
@@ -57,6 +59,103 @@ def step_up(x: Any, step: int) -> int:
...
+    result = timing_groups.resolve_groups(groups, pooled, base, _eval)
+    return TimingProfile(
+        timeLimit=result.base_time_limit,
+        formula=formula,
+        timeLimitPerLanguage=result.time_limit_per_language,
+        groups=result.reports,
+    )
+
+
+async def _prompt_repartition(
```
rsalesc: I don't like this is being asked with sequential but separate questionary texts. I'd really love for this to be an interactive UI where

1. every language is shown at the same time with `[X] language`.
2. X is either empty or a positive integer, denoting the group number this language is part of
3. you can select up/down these languages, like you would with the questionary choice feature
4. when selecting each entry, you can hit 0-9 to put them into a language group.

Please do this even if you have to implement a separate UI library for it.

tests/rbx/box/test_limits_table.py
```diff
@@ -0,0 +1,97 @@
...
+        modifiers={'java': LimitModifiers(timeMultiplier=2.0)},
+    )
+    rows = build_limits_table_rows(profile)
+    java_row = next(r for r in rows if r.languages == 'java')
+    assert java_row.time_limit_ms == 2000
+    assert '2.0' in java_row.source
+
+
+def test_render_limits_table_highlights_defaulted(capsys):
+    from rbx.box.limits_info import render_limits_table
```
rsalesc, line 82: This rendered table is too grey, please let's add a bit of color to it to align with other tables, which are usually rendered in white, and with proper colors to highlight things that make sense.

rbx/box/timing.py
rsalesc: Done in 9510227. Replaced the per-language `questionary.text` loop with a single interactive prompt_toolkit screen (`rbx/box/timing_group_picker.py`): all languages are listed at once as `[N] language` (empty box = own group), navigable with ↑/↓ or j/k, and pressing 0-9 assigns the highlighted language to that group; Enter confirms, q/Ctrl-C cancels. The assignment logic lives in a pure `GroupPickerState`; key bindings + navigation + confirm/cancel are covered by piped-input tests (no new dependency — prompt_toolkit is already transitive via questionary).

tests/rbx/box/test_limits_table.py
rsalesc: Done in 5a98a0b. The greyness came from the console's base `info` style; the table now sets explicit styles — bold-white title/header, blue (`item`) languages, bold-white time limits, white solutions/source — with meaningful highlights: estimated sources in green (`success`), ×multiplier sources in blue (`item`), and DEFAULTED rows fully in `warning` (yellow) with ⚠. Refactored into a testable `build_limits_table` and added a test asserting the columns carry non-grey styles.

PR #514, opened 2026-06-03
feat(timing): force relative time limits in the rbx time picker (#512)

### LINKED ISSUE #512: Brainstorm a way of interactively allowing `rbx time` users to specify time limit relative to another language
After #499 , users have a way to specify that the time limit of a language group X should be At+B the time limit t of another language Y. Also, the user has the capability of reassigning groups during `rbx time` execution.

When the user modifies a group, they completely lose the ability to choose a relative parameter. The user should have a way to FORCE a relative assignment to happen, and to edit its parameters (A and B) in the prompt picker for the groups.

I'm not sure what would be a good UX for that, but something with hotkeys and custom prompts should do the job.

rsalesc:
Closes #512.

## What

Lets `rbx time` users **force** a language group's time limit to be computed relative to another group as `A·t + B` (`multiplier × reference + increment`), directly in the interactive picker. Unlike env `whenEmpty` (which only fires for groups with no solutions), a forced relative **always overrides** the group's measured estimate — the live preview shows solution counts so the choice is informed.

## Picker UX

- **`r`** on a language opens an inline editor for its group: `relative-to: [ref]  A:[..]  B:[..]`. `Tab` cycles the focused field (ref → A → B), `←/→` cycles the reference group (including a **base estimate** option), type to set A/B, `Enter` commits, `Esc` cancels, `c` clears.
- A persistent `→ ref ×A +B` annotation marks relative groups in the list.
- **`R`** resets the whole grouping + relative rules back to what `env.rbx.yml` defines.
- The live preview validates references inline (self-reference / cycles) and blocks confirm.

## Design / approach

- New `ResolvedGroup.forced_relative` takes top priority in `resolve_groups` (forced → measured timings → env `whenEmpty` empty-only → base). The env schema is untouched; the forced spec lives only on the resolved layer.
- `partition_from_assignment` **drops the old env-crossing** (it no longer re-derives `whenEmpty` by matching membership). Instead env `whenEmpty` is **seeded into picker state at init**, only for groups with no solutions. After init, the picker's `{numbers, relatives}` is the single source of truth — membership edits are the user's responsibility, and `R` recovers the env grouping.
- The picker return grew from a raw dict to a `GroupAssignment{numbers, relatives}`; relatives are keyed by a stable group-key (`g{N}`/`s:{lang}`/`leftover`) and pruned when a group disappears.
- Resolved profiles store concrete numbers, so nothing round-trips to `env.rbx.yml`.

Design + plan docs: `docs/plans/2026-06-03-rbx-time-forced-relative-design.md`, `docs/plans/2026-06-03-rbx-time-forced-relative-plan.md`.

## Testing

- Full timing suite green (118 tests): `test_timing_groups`, `test_timing_group_picker`, `test_timing`, `test_timing_estimation`, `test_timing_preview`, `test_limits_table`.
- Covers: forced relative winning over a group's own timings (unit **and** end-to-end through `build_timing_profile`), base-estimate reference, validation (self/cycle) reachable from the picker path, editor key routing (digit routing, field cycling, ref cycling, clear/cancel/commit), reset-to-env, init-seeding only for empty groups, and prune of orphaned specs.
- Broader box suite: no new regressions (remaining failures are pre-existing, unrelated C++/sandbox/docker/walltime env issues).

### DIFF
--- a/docs/setters/profiling/index.md
+++ b/docs/setters/profiling/index.md
@@ -196,6 +196,22 @@ The picker is **prepopulated from `env.rbx.yml`**: any groups you configure ther
 preselected. After estimation, a per-group table is printed showing the estimated time limit
 for each group.
 
+### Forcing a relative limit
+
+While in the picker you can **force** a group's time limit to be computed relative to another
+group instead of from its own measured timings:
+
+- Press <kbd>r</kbd> on a language to open an inline editor and force its group's limit to
+  `A·t + B`, where `t` is the reference's limit, `A` is the multiplier, and `B` is an optional
+  increment in milliseconds. The reference can be another group or the **base estimate**. In the
+  editor, <kbd>Tab</kbd> cycles between the reference, `A`, and `B`; type to set `A`/`B`;
+  <kbd>Enter</kbd> commits, <kbd>Esc</kbd> cancels, and <kbd>c</kbd> clears the rule.
+- Press <kbd>R</kbd> to reset the whole grouping and all relative rules back to what
+  `env.rbx.yml` defines.
+
+Unlike the env `whenEmpty` fallback (which only applies to groups that have **no** solutions),
+a forced relative **always overrides** the group's measured estimate.
+
 Press <kbd>Enter</kbd> to confirm, or pass `--auto` to skip the prompt and use the configured
 env groups as-is.
 
--- a/docs/setters/profiling/index.md
+++ b/docs/setters/profiling/index.md
@@ -201,11 +201,13 @@ for each group.
 While in the picker you can **force** a group's time limit to be computed relative to another
 group instead of from its own measured timings:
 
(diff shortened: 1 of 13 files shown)


docs/setters/profiling/index.md
```diff
@@ -196,6 +196,24 @@ The picker is **prepopulated from `env.rbx.yml`**: any groups you configure ther
...
 
+### Forcing a relative limit
+
+While in the picker you can **force** a group's time limit to be computed relative to another
+group instead of from its own measured timings:
+
+- Press <kbd>r</kbd> on a language to open an inline editor and derive its group's limit from
+  another group as `multiplier × reference + increment` (the formula is shown in the editor).
+  The reference can be another group or the **base estimate**; the increment is an optional
+  constant in milliseconds. In the editor, <kbd>Tab</kbd> switches the focused field
```
rsalesc, line 207: no need to detail something that can be seen while running hte command

PR #515, opened 2026-06-03
feat(presets): make the default preset lean (#413)

### LINKED ISSUE #413: Change default preset to be lean
At the moment, the default preset has a bunch of example solutions, a `testplan.py` and various other things that are usually not required, and that in practice people will likely delete.

The initial idea was for the preset to serve as a reference for everything that is possible, but this harms more than do good.

We can make the default problem preset be really basic, and have more documentation around what are the possibilities, so chief setters don't have to go around deleting things.

Additionally, we can provide a full problem example during the contest creation if this helps.

rsalesc:
## Summary

Closes #413. Slims the default preset to a bare, buildable A+B problem and moves "everything you can do" into in-preset docs, so chief setters no longer start by deleting scaffolding.

- **Lean problem**: `problem.rbx.yml` keeps all solution *outcome declarations* (`ac-/wa-/tle-/…`) but ships only `sols/main.cpp`; drops `stresses`, `unitTests`, and the `program-random` group. `vars` retained (validator/statement read `N`).
- **Reorganized layout**: `documents/` (statement + assets + samples), `tests/` (commented-only `testplan.txt` + `gens/gen.cpp`), `sols/`; validator/checker/headers at root. The `random` testcase group is renamed `testplan`. Contest folder `statement/` → `documents/` too, with `preset.rbx.yml` tracking retargeted.
- **In-preset docs**: a short `problem/README.md` (ships into every created problem) plus brief doc-pointer comments, all linking to the rbx docs.
- **e2e**: new `seed_from_preset` harness primitive seeds a scenario's temp package from the *real* preset at run time (dereferencing the `shared/` symlinks, skipping cruft), so the `default-preset` fixture exercises the live preset instead of a static copy that could drift.
- **Setter docs updated** for the new layout: `first-steps.md` (tree + reframed so the reader *creates* `sols/wa-overflow.cpp`), `stress-testing-walkthrough.md`, `verification/validators.md`, `presets/index.md`.

Design + plan: `docs/plans/2026-06-03-lean-default-preset-design.md`, `docs/plans/2026-06-03-lean-default-preset.md`.

**Deferred to follow-up issues**: a complete unlinked example problem inside the contest dir; a dedicated `complete-problem` e2e fixture (WA/unit/packaging coverage); migrating generic `statement/` example paths in cheatsheet/statements docs to `documents/`. **Out of scope**: cruft cleanup (`.box/`, `build/`, `.DS_Store`).

## Test Plan

- [x] Preset builds standalone (`rbx build`), `sols/main.cpp` judges AC (`rbx run`), statement renders to PDF (`rbx statements build`)
- [x] e2e suite: 142 passed (lone failure is the pre-existing flaky `test_collects_scenarios`, untouched by this branch); new `seed_from_preset` unit tests pass
- [x] preset/creation/BOCA-default-preset tests: 108 passed
- [x] broad box suite: 2202 passed — remaining failures are confined to `checkers_test.py` (pre-existing local g++ compilation env issue, not touched here)
- [x] `ruff check` / `ruff format --check` clean; `mkdocs build` clean


rbx/resources/presets/default/problem/problem.rbx.yml
```diff
@@ -1,5 +1,7 @@
 ---
 # yaml-language-server: $schema=https://rsalesc.github.io/rbx/schemas/Package.json
+# Minimal A+B problem. See the README in this preset, and the rbx docs at
```
rsalesc: Let's keep this lean. This means one core principle should be: let's not add things that are just there for the person using the preset to create the problem to remove.

Documenting this is just a minimal A+B problem brings nothing to the table, and rather just adds something the user must remove.

rbx/resources/presets/default/problem/README.md
```diff
@@ -0,0 +1,45 @@
+# rbx default problem
+
+A minimal, ready-to-build A+B problem. Run `rbx build` to generate the testset
```
rsalesc: Same as I mentioned in the other comment, assume this is a ready-to-increment problem the users shouldn't have to go through the code base finding things to delete.

rbx/resources/presets/default/problem/problem.rbx.yml
rsalesc: Done in 6467871. Removed the `# Minimal A+B problem…` header. Applied the same principle across the rest of the preset: dropped the `// … Docs:` pointer comments in `gen.cpp` and `validator.cpp` (reverted both to their original commentless form) and the prose/Docs header in `tests/testplan.txt` (it now ships just the two commented example calls). Left the pre-existing field-level inline hints (units, `download checker` tip, etc.) since they predate this PR and aren't pure delete-bait — happy to do a separate pass on those if you want them gone too.

rbx/resources/presets/default/problem/README.md
rsalesc: Done in 6467871 — deleted the in-problem README entirely. Since it ships into every created problem, any README is exactly the kind of thing a setter has to delete. "What you can add" discovery now lives only in the rbx docs site (first-steps + the per-feature setter pages this PR already updated), which is where issue #413's "documentation around the possibilities" belongs. Updated the first-steps directory tree to drop the README node.

rsalesc: Re-added in a73e545. I'd over-applied the principle and deleted it outright — that was wrong. It's back as a durable reference (package layout + the solution-prefix convention + doc links), scrubbed of the A+B-specific framing, so it passes your survives-edits test and is something the setter keeps rather than deletes.

Issue #522, opened 2026-06-05
Rethink complilation and execution working directory in sandbox


rsalesc:
Nowadays, when we run a program, we create a temporary directory within the sandbox, we copy the program that should be compiled (in case of compilation) or the compiled program that should be executed (in case of execution) to the sandbox, and run them.

Besides that, we declare a few files that should be either copied in into this temp directory, or copied out from this directory after the program finishes. These are the grading artifacts. This is a very important feature to ensure we can cache the compilations and executions properly. Basically, we ensure that every artifact used by the program is declared in sort a "manifest" so we can properly track them in the cache to evict when necessary.

This works well, but is cumbersome for some reasons:

1. Programs must declare in a `compilationFiles` field in `problem.rbx.yml` files that should be included during the compilation. In fact, there isn't even a similar mechanism for "execution files", so it's virtually impossible to fulfill this use case nowadays. These files are included, and tracked by the cache/manifest properly, but, although most of the times necessary, are a bit cumbersome to declare.
2. The temporary directory doesn't reflect the problem package directory structure. Let's suppose there's a C++ program in `gens/gen.cpp`, and a `lib.h` at the root of the package. Ideally, the C++ program should #include "../lib.h" to use the library. It's essential that we somehow mirror this directory structure in the temp directory, and make sure we compile/run the program with the current `cwd` to ensure proper functioning.

NOTE: it's important to understand how things would work with (2) nowadays, we should research.

---

We should be able to brainstorm solutions for this, in a way that we minimize the number of changes necessary to ensure the tool won't break as this is a pretty tricky migration.

(1) in particular is a smaller priority, but we definitely should support both "compilationFiles" and "executionFiles" for every `CodeItem`. We should also have an interface that is language specific (triggered by checks such as "is_cxx_lang" and "is_python_lang") to *automatically expand* "compilationFiles" and "executionFiles" based on #include / relative import statements found in those files.

PR #523, opened 2026-06-05
Mirror package directory structure in sandbox (Phase 1, #522)

rsalesc:
## Summary

Phase 1 of #522 — make the sandbox working directory mirror the package layout, so a source in a subdirectory can include files elsewhere in the package via relative paths.

- **Source mirroring** (`code.py`): `_get_code_variables` now sets `source` to the package-relative path, so `gens/gen.cpp` is compiled at `gens/gen.cpp` (driven through the existing `{source}` → `FileMapping` plumbing). The `cwd` stays the sandbox root, which now faithfully mirrors the package root.
- **`compilationFiles` repurposed** (`package.py`): files land at their package-relative path and may live anywhere **under the package root** (the old "must be under the code's folder" restriction is lifted), enabling `#include "../lib.h"`. New `get_relative_source_path` resolver (with a basename fallback for remote/external paths).
- **Builtin headers** (`download.py`): testlib/jngen/tgen/rbx are placed in the **source's own directory** so quoted `#include "testlib.h"` resolves without `-I.`; precompiled `.gch` files land beside them automatically.
- **Docs**: corrected the user-facing `compilationFiles` field description.

**Back-compat:** flat packages are byte-identical (package-relative path == basename at the root); Java/Kotlin are untouched (they map via `{javaClass}`/`Main.kt`, not `{source}`); the `<bits/stdc++.h>` `-I.` mechanism is unaffected. Net production change is ~32 lines across 4 files.

**Deferred to Phase 2** (per the issue's stated priority): `executionFiles` on every `CodeItem`, and default-on, recursive, quoted-only auto-expansion of `#include "..."` (C++) and relative/sibling imports (Python).

Design & plan: `docs/plans/2026-06-05-sandbox-working-directory-design.md` and `docs/plans/2026-06-05-sandbox-working-directory.md`.

## Test plan

- [x] Unit wiring tests (`code_compile_test.py`) + 4 real-compile integration tests (`code_compile_integration_test.py`): `../lib.h` parent include, subdir testlib resolution, flat-source regression, nested Python generator run — **44 passing locally**, ruff clean.
- [x] Verified **zero regressions**: the broad box suite's failures on the dev machine (real-compile under clang + rich-rendering string assertions) reproduce **identically at the base commit** `d6d94f5`, so they are pre-existing and unrelated.
- [ ] CI green on a Linux/g++ runner (note: this branch was developed on macOS/clang, where some unrelated suites fail pre-existingly).

### DIFF
--- a/rbx/box/code.py
+++ b/rbx/box/code.py
@@ -670,12 +670,15 @@ async def compile_item(
             with profiling.Profiler('code.precompile'):
                 precompilation_inputs = []
                 for input in artifacts.inputs:
-                    # Precompile the tool-injected headers in __internal__/; only
-                    # those are placed there, so no name allow-list is needed.
+                    # Precompile the static tool-injected headers in __internal__/.
+                    # rbx.h is skipped: it is regenerated per problem (it embeds the
+                    # package vars), so precompiling it yields no cross-problem cache
+                    # reuse.
                     if (
                         input.src is not None
                         and input.src.suffix == '.h'
                         and steps.is_internal_path(input.dest)
+                        and input.dest.name != 'rbx.h'
                     ):
                         precompilation_inputs.append(
                             await _precompile_header(
--- a/rbx/box/code.py
+++ b/rbx/box/code.py
@@ -217,7 +217,10 @@ def _get_java_class_name(compilable_path: pathlib.Path) -> Optional[str]:
 
 
 def _get_code_variables(code: CodeItem, language: str) -> dict[str, Any]:
-    res = {'source': code.path.name, 'language': language}
+    res = {
+        'source': package.get_relative_source_path(code).as_posix(),
+        'language': language,
+    }
     java_klass = _get_java_class_name(code.path)
     if java_klass is not None:
         res['javaClass'] = java_klass
--- a/rbx/box/code.py
+++ b/rbx/box/code.py
@@ -670,12 +670,12 @@ async def compile_item(
             with profiling.Profiler('code.precompile'):
                 precompilation_inputs = []
                 for input in artifacts.inputs:
+                    # Precompile the tool-injected headers in __internal__/; only
+                    # those are placed there, so no name allow-list is needed.
                     if (
                         input.src is not None
                         and input.src.suffix == '.h'
-                        and input.dest.is_relative_to(steps.INTERNAL_DIR)
-                        and input.dest.name
-                        in ['stdc++.h', 'jngen.h', 'tgen.h', 'testlib.h']
+                        and steps.is_internal_path(input.dest)
                     ):
                         precompilation_inputs.append(
                             await _precompile_header(
--- a/rbx/box/code.py
+++ b/rbx/box/code.py
@@ -652,12 +652,15 @@ async def compile_item(
         for input in artifacts.inputs:
             await _ignore_warning_in_cxx_input(input)
 
-        # Add system bits/stdc++.h to the compilation.
+        # Add system bits/stdc++.h to the compilation. It lives in the reserved
+        # __internal__/ dir (alongside the builtin headers) and is exposed via
+        # -I__internal__ so <bits/stdc++.h> resolves there without putting the
+        # mirrored package root on the include path.
         bits_artifact = maybe_get_bits_stdcpp_for_commands(commands)
         if bits_artifact is not None:
             artifacts.inputs.append(bits_artifact)
             commands = [
-                command + ' -I.'
+                command + f' -I{steps.INTERNAL_DIR}'
                 for command in commands
                 if is_cxx_command(get_exe_from_command(command))
             ]
@@ -670,6 +673,7 @@ async def compile_item(
                     if (
                         input.src is not None
                         and input.src.suffix == '.h'
+                        and input.dest.is_relative_to(steps.INTERNAL_DIR)
                         and input.dest.name
                         in ['stdc++.h', 'jngen.h', 'tgen.h', 'testlib.h']
                     ):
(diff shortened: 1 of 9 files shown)


rsalesc:
## Adversarial review + follow-up changes (via Claude Code)

Recording the review and the changes pushed in `dc4edc2` so the context isn't lost.

### Review findings (against the prior state of this PR)

**🔴 Confirmed back-compat regression (now fixed).** A C++ source in a subdirectory that included a companion header with **angle brackets** — `#include <lib.h>` — compiled on `main` (`d6d94f5`) but **failed on the PR branch**. Verified with a differential real-compile test:
- base `d6d94f5`: compiles ✅
- PR (before this change): `'lib.h' file not found with <angled> include; use "quotes" instead` ❌

Mechanism: the PR mirrored the companion to `gens/lib.h` but kept `-I.` (sandbox root only) for the bits/stdc++.h shim. Angle-bracket resolution searches only `-I` dirs, so the subdir companion was unreachable. (Correction to my first pass: `-I.` is added on essentially **every** C++ compile, not just macOS/clang — `maybe_get_bits_stdcpp_for_commands` also copies the *system* `bits/stdc++.h` in on Linux/g++ — so this hit the CI path too.)

**🟡 Builtin-header shadowing.** With builtins copied beside the source, the `download.py` dedup (`dest == artifact.dest`) could miss when a user's own `testlib.h` sat at a different package-relative path than the source dir, silently shadowing it.

Other items (non-blocking, noted for the record): the design doc's "one-time **global** recompile" overstated impact (flat packages were byte-identical); the Python `sys.path[0]` shift for mirrored generators (no companion files exist at runtime yet, so no concrete break).

### What changed in `dc4edc2`

Reworked the header-injection mechanism into a single reserved sandbox dir, which fixes the regression and dissolves the shadowing:

- **`__internal__/` include root.** Builtins (testlib/jngen/tgen/rbx) **and** `bits/stdc++.h` now all land in `__internal__/`, exposed via one **`-I__internal__`** flag (replacing `-I.`). Because a *quoted* `#include "x"` searches the source dir first and then the `-I` dirs, `#include "testlib.h"` resolves from `__internal__/` for **any** source location — no per-source placement needed, and the mirrored package root is no longer on the include path.
- **Removed the `download.py` dedup.** Builtins in `__internal__/` can't collide with package-relative `compilationFiles`, so the check is dead weight. Per discussion, no suppression is needed: a user's own header still wins when declared **beside the source** (source-relative resolution precedes the `-I__internal__` fallback); otherwise the builtin is the fallback.
- **Tightened the precompile filter** to only headers under `__internal__/`, so a user `compilationFile` named like a builtin isn't needlessly precompiled. The `.gch` follows its header into `__internal__/`.
- **Source mirroring (`_get_code_variables`) and the `compilationFiles` repurpose are unchanged** — that part was solid.

**Accepted behavior change:** angle-bracket includes of *project* headers (`#include <lib.h>`) no longer resolve — quoted includes are the supported form (testlib/jngen/tgen all use quotes; `<...>` should mean system/stdlib). This is intentional and consistent; it also drops the incidental flat-layout `-I.` support.

The design doc (`docs/plans/2026-06-05-sandbox-working-directory-design.md`) §3–§4 and the migration notes were updated to describe this approach.

### Tests

- Unit wiring (`code_compile_test.py`): builtins/bits assert `__internal__/...`, command asserts `-I__internal__`, precompile targets `__internal__/`, and a new test that a user header named like a builtin elsewhere is **not** precompiled.
- Real-compile integration (`code_compile_integration_test.py`): subdir `#include "testlib.h"`, subdir `#include <bits/stdc++.h>`, `../lib.h` parent include, flat source, nested Python generator, and a new test proving a declared user `testlib.h` beside the source shadows the builtin.
- **55 passing locally**, ruff clean. The one failing local test (`code_run_test.py::...test_stack_limit_check_..._with_hard_limit`) fails **identically on `d6d94f5`** — a pre-existing macOS ulimit issue, unrelated to this change.

rbx/box/code.py
```diff
@@ -667,6 +673,7 @@ async def compile_item(
                     if (
                         input.src is not None
                         and input.src.suffix == '.h'
+                        and input.dest.is_relative_to(steps.INTERNAL_DIR)
```
rsalesc: please create a helper in steps.py to identify whether a path (inside the sandbox) is an internal dir (very similar to this line)

rbx/box/code.py
```diff
@@ -667,6 +673,7 @@ async def compile_item(
...
                         and input.dest.name
                         in ['stdc++.h', 'jngen.h', 'tgen.h', 'testlib.h']
```
rsalesc: with the internal dir check, this isn't necessary anymore :)

only such files are placed in there

rbx/box/code.py
rsalesc: Done in `be7ae11` — added `steps.is_internal_path(path)` (sits right under `INTERNAL_DIR`) and the precompile loop now uses it instead of the inline `is_relative_to`.

rbx/box/code.py
rsalesc: Done in `be7ae11` — dropped the `dest.name in [...]` allow-list; the `is_internal_path` check is the sole filter now.

One heads-up: the allow-list silently excluded `rbx.h`, so removing it means `rbx.h` now gets precompiled too. It compiles fine (verified by the real-compile integration suite), so no correctness issue. The only nuance is cost: `rbx.h` is regenerated per problem (it embeds the package vars), so unlike the static testlib/jngen/tgen/bits headers its precompile cache key changes per problem — it precompiles once per unique `rbx.h` and is reused across that problem's C++ compiles, but the `.gch` is only actually consumed when a source includes `rbx.h` first. If you’d rather not pay that, say the word and I’ll gate it (e.g. skip headers whose content is package-derived). Otherwise the simpler dir-only filter stands.

rsalesc: Followed up in `7b64672`: `rbx.h` is now excluded from precompilation (it is regenerated per problem, so no cross-problem cache reuse). The precompile filter is `is_internal_path(dest) and dest.name != "rbx.h"`, so testlib/jngen/tgen/bits are precompiled and rbx.h is not.

Issue #525, opened 2026-06-06
Packaging: flatten sources and rewrite cross-directory #include "..." for flat/inline judges (Polygon, BOCA)


rsalesc:
Follow-up to #522 / #523 (Phase 1). Phase 1 made the local sandbox mirror the package layout, so a source in a subdirectory can `#include "../lib.h"` (or `#include "common/x.h"`) and compile/run locally. The export packagers were **not** updated and cannot represent these layouts, so a package can build + verify green locally yet be silently broken once exported.

## Problem

Packaging re-exports source files to the target judge, which compiles them in *its own* environment (`rbx/box/packaging/CLAUDE.md`). Two of the targets have no notion of directories:

- **Polygon is strictly flat** — one flat file namespace, no directories whatsoever.
  - Offline `problem.zip` (`polygon/packager.py:197`): the checker/interactor are compiled by Polygon in `files/`, with only `testlib.h`/`rbx.h` beside them (`_get_files`, lines 159–163).
  - API upload (`polygon/upload.py`): generators/validators/solutions/checker/interactor are uploaded **by basename** into the flat file space.
- **BOCA** (`boca/packager.py`, `resources/packagers/boca/checker.sh`): the checker (and interactor) source is embedded **inline** in the compile script and compiled flat with only `testlib.h`/`rbx.h` written beside it (`g++ {{rbxFlags}} checker.cpp -o checker.exe` — no `-I`, no other files).

So any exported source that uses a cross-directory quoted include (`#include "../lib.h"`, `#include "subdir/x.h"`) fails to compile on the target:
- Polygon offline: checker / interactor.
- Polygon upload: any generator / validator / solution / checker / interactor.
- BOCA: checker / interactor.

There is **no** flatten / rename / include-rewrite step anywhere under `rbx/box/packaging/`. This becomes the *default* failure mode once #524 (Phase 2) ships default-on auto-expansion of `#include "..."`.

## Scope

When exporting/uploading a source whose mirrored layout is not flat:
1. Collect the source plus its transitive quoted-include dependencies (reuse the include scanner from #524).
2. Copy each into the flat export namespace under a unique, collision-free name (coordinate with #523-follow-up "basename collisions" issue).
3. Rewrite the quoted `#include "..."` directives in the exported copies to point at the flattened names.

[...]

PR #528, opened 2026-06-06
feat: sandbox dir mirroring Phase 2 — executionFiles + auto-expansion (#524)

### LINKED ISSUE #524: Sandbox dir mirroring Phase 2: executionFiles + auto-expansion of includes/imports
Follow-up to #522 (Phase 1 shipped in #523), covering the scope deliberately deferred per the issue's stated priority. See the design doc `docs/plans/2026-06-05-sandbox-working-directory-design.md` (§ "Phase 2") and the implementation plan `docs/plans/2026-06-05-sandbox-working-directory.md`.

## Context

Phase 1 made the sandbox working directory mirror the package layout: sources compile at their package-relative path, `compilationFiles` land package-relative (the under-code-folder restriction was lifted, enabling `#include "../lib.h"`), and builtin headers (testlib/jngen/tgen/rbx) are placed in the source's own directory. Manual declaration is currently required to bring extra files into the mirror.

## Scope

1. **`executionFiles` on every `CodeItem`** — the runtime equivalent of `compilationFiles`. Declared files are mirrored into the sandbox for *execution* (today there is no execution-time equivalent at all). Feeds the run-side `GradingArtifacts` in `_prepare_run`/`run_item`. For a flat package this is a no-op; for a subdir source the file lands at its package-relative path.

2. **Default-on, recursive, quoted-only auto-expansion** of dependencies, language-specific (gated by `is_cxx`/`is_python`-style checks):
   - **C++**: parse `#include "..."` (ignore `<...>` system includes), follow transitively, auto-add the discovered files as compilation files (and execution files where relevant).
   - **Python**: relative/sibling imports (`from . import x`, `from ..pkg import y`, `import sibling`), follow transitively, auto-add as execution files.
   - Manual `compilationFiles` / `executionFiles` remain as an override / escape hatch.

## Known Phase-1 limitations to revisit here

- **Builtins only in the source's directory.** A header reached cross-directory (via `../`) that *itself* does `#include "testlib.h"` won't find the builtin (it only sits beside the primary source). Intended resolution: deprecate auto-placed builtin libraries (testlib/jngen/tgen/rbx) in favor of explicit declaration + auto-expansion.
- **Dedup cosmetic shadowing (minor).** If a user declares a *root* builtin in `compilationFiles` for a *subdir* source (a combination newly accepted in Phase 1), both a root copy and a builtin-derived source-dir copy are added. The compile still succeeds (source-dir copy wins for the quoted include); worth a dedup tidy or a one-line comment.

## Testing

- Auto-expansion unit tests: nested/transitive `#include "..."`, ignore `<...>`, Python relative + sibling imports; cycle safety.
- Integration (teeth-having):
  - Subdir C++ source whose `#include "../lib.h"` is **auto-discovered** (no manual `compilationFiles`).
  - Subdir Python source importing a **sibling module** runs correctly end-to-end — this is the execution-mirroring test Phase 1 could only smoke-test (Phase 1 added a nested-Python-generator run guard, but without imports it passes under the old flat layout too).

rsalesc:
## Summary

Implements **Phase 2** of sandbox directory mirroring (#524), follow-up to #522/#523. Adds an `executionFiles` field and **default-on, transitive, quoted-only auto-expansion** of dependencies, built on a new per-language dependency interface that is deliberately shaped so #525 (packaging flatten/rewrite) can reuse it.

Design & plan: `docs/plans/2026-06-06-sandbox-working-directory-phase2-design.md` and `…-phase2.md`.

## What's in here

**New `rbx/box/dependencies/` package** (mirrors `linters/` — registry + self-registering per-language modules):
- `scanner.py` — `DependencyKind`, `Reference(spelling, target)`, `DependencyScanner` ABC + registry.
- `cpp.py` — **tree-sitter-cpp** scanner: `references()` (quoted `#include "…"` only; `<...>`, comments, and builtins → unresolved) and `rewrite()` (span-based, byte-preserving, `can_rewrite=True`).
- `python.py` — `ast` scanner: relative/sibling imports (`from .`, `from ..pkg`, `import sibling`), plus intermediate `__init__.py` markers for deep dotted imports.
- `graph.py` — `expand(code, require_kind=…)` → `DependencyGraph` (cycle-safe; `files()` excludes root; per-file edges exposed for #525).

**Wiring:**
- `CodeItem.executionFiles` + `package.get_execution_files` (shares one declared-files helper with `get_compilation_files`), propagated through the `*WithDigest.create` factories.
- `compile_item` unions C++ `#include` deps into compile inputs (additive with manual `compilationFiles`).
- `_prepare_run` mirrors manual `executionFiles` + auto-discovered Python deps into run inputs.

## Notable: a real runtime bug fixed

Mirroring the sibling module into the sandbox was **necessary but not sufficient** for Python. The entry script is materialized as a symlink into the content-addressed cache, so Python ≥3.11 resolves `sys.path[0]` to the cache dir (via the symlink's realpath), **not** the mirrored source dir — so `import sibling` failed with `ModuleNotFoundError` even with the sibling sitting next to the script. Fix: set `PYTHONPATH` to the mirrored source dir for interpreted runs (cache-key-correct via `set_env`). This was latent since Phase 1; #524 is the first feature to exercise runtime sibling imports end-to-end.

## #525 — interface baked, not built

The `DependencyGraph` (per-file `spelling → target` edges, including the root) + the C++ `rewrite(text, rename)` primitive are delivered and unit-tested so #525 can flatten + rewrite cross-directory includes for Polygon/BOCA. **No `packaging/` code is touched** in this PR.

## Out of scope (deliberately)

- All `packaging/` work (that's #525) and its interim guardrail.
- Deprecating auto-injected builtins (already moot: Phase 1's `__internal__/` revision resolves cross-dir builtin includes).
- Python source flattening / `rewrite` (gated behind `can_rewrite=False`).

## Testing

- **Unit:** C++ `references`/`rewrite` (nested/transitive, ignore `<...>`, comment-safe, builtin→None); Python relative/sibling/deep imports + intermediate `__init__.py`; `expand` cycle-safety, refuse-outside-root, `None` for unhandled languages, `require_kind` short-circuit; `executionFiles` schema + factory propagation.
- **Integration (real toolchain):** subdir C++ `#include "../lib.h"` **auto-discovered** (no manual `compilationFiles`) compiles; subdir Python **sibling import runs end-to-end** (the teeth-having test Phase 1 could only smoke-test); flat-package regression.
- 63 affected tests pass; lint clean; no import cycle. Pre-existing environmental failures (validators/generators/stack-limit) verified identical to `main` — no regressions.
- An adversarial code-review pass flagged two items, both addressed: `expand` now short-circuits the tree-sitter walk for C++ on the per-run hot path, and deep `import a.b.c` ships intermediate `__init__.py`.

Closes #524.

### DIFF
--- a/rbx/box/code.py
+++ b/rbx/box/code.py
@@ -419,6 +419,24 @@ async def _prepare_run(
     if outputs:
         artifacts.outputs.extend(outputs)
 
+    # Mirror manual + auto-discovered execution files at their package-relative path.
+    # C++ contributes none here (its deps are compile-time); Python siblings land via
+    # auto-expansion. Manual executionFiles remain the escape hatch for all languages.
+    from rbx.box.dependencies import graph as deps_graph
+    from rbx.box.dependencies.scanner import DependencyKind
+
+    exec_dests = {input.dest for input in artifacts.inputs}
+    for src, dest in package.get_execution_files(code):
+        if dest not in exec_dests:
+            artifacts.inputs.append(GradingFileInput(src=src, dest=dest))
+            exec_dests.add(dest)
+    dep_graph = deps_graph.expand(code)
+    if dep_graph is not None and DependencyKind.EXECUTION in dep_graph.kinds:
+        for dep in dep_graph.files():
+            if dep not in exec_dests:
+                artifacts.inputs.append(GradingFileInput(src=dep, dest=dep))
+                exec_dests.add(dep)
+
     return PreparedRun(
         command=command,
         sandbox_params=sandbox_params,
--- a/rbx/box/code.py
+++ b/rbx/box/code.py
@@ -29,6 +29,7 @@ from rbx.box.environment import (
     get_mapped_command,
     get_mapped_commands,
     get_sandbox_params_from_config,
+    is_interpreted,
     merge_execution_configs,
 )
 from rbx.box.formatting import get_formatted_memory
@@ -384,11 +385,18 @@ async def _prepare_run(
         command = shlex.join(splitted_command)
 
     artifacts = GradingArtifacts()
+    # Interpreted entry scripts must be a real copy, not a cache symlink: the
+    # interpreter resolves its module search root from the script's realpath, which
+    # for a symlink points back into the content cache instead of the mirrored
+    # source dir. Copying keeps that realpath inside the sandbox so sibling imports
+    # resolve naturally.
+    interpreted = is_interpreted(language, solution=isinstance(code, Solution))
     artifacts.inputs.append(
         GradingFileInput(
             **executable.expand(),
             dest=PosixPath(file_mapping.executable),
             executable=True,
+            symlink=not interpreted,
         )
     )
     if stdin is not None:
@@ -431,20 +439,11 @@ async def _prepare_run(
             artifacts.inputs.append(GradingFileInput(src=src, dest=dest))
             exec_dests.add(dest)
     # Only interpreted languages (EXECUTION-kind scanner) contribute here; C++ and
-    # other compiled languages short-circuit inside expand() without walking.
+    # other compiled languages short-circuit inside expand() without walking. The
+    # interpreted entry script is copied (not symlinked) above, so its mirrored
+    # directory is the interpreter's module search root and these siblings resolve.
     dep_graph = deps_graph.expand(code, require_kind=DependencyKind.EXECUTION)
     if dep_graph is not None:
-        # The entry script is symlinked from the content-addressed cache, so the
-        # interpreter resolves its module search root to the cache dir (via the
-        # symlink's realpath) instead of the mirrored source dir. Put the mirrored
-        # source dir back on PYTHONPATH so sibling imports resolve, matching how
-        # `python3 <dir>/script.py` normally behaves. This overrides any PYTHONPATH
-        # already in set_env (none is, today); the host PYTHONPATH is not inherited.
-        source_dir = dep_graph.root.parent.as_posix()
-        existing_pythonpath = sandbox_params.set_env.get('PYTHONPATH')
-        sandbox_params.set_env['PYTHONPATH'] = (
-            f'{source_dir}:{existing_pythonpath}' if existing_pythonpath else source_dir
-        )
         for dep in dep_graph.files():
             if dep not in exec_dests:
                 artifacts.inputs.append(GradingFileInput(src=dep, dest=dep))
--- a/rbx/box/code.py
+++ b/rbx/box/code.py
@@ -628,6 +628,19 @@ async def compile_item(
             for src, dest in package.get_compilation_files(code)
         )
 
+        # Auto-expand transitive quoted #include "..." dependencies (default-on,
+        # additive). Manual compilationFiles remain the escape hatch.
+        from rbx.box.dependencies import graph as deps_graph
+        from rbx.box.dependencies.scanner import DependencyKind
+
+        dep_graph = deps_graph.expand(code)
+        if dep_graph is not None and DependencyKind.COMPILATION in dep_graph.kinds:
+            existing = {input.dest for input in artifacts.inputs}
+            for dep in dep_graph.files():
+                if dep not in existing:
+                    artifacts.inputs.append(GradingFileInput(src=dep, dest=dep))
+                    existing.add(dep)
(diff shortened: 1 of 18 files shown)


rbx/box/code.py
```diff
@@ -419,6 +419,37 @@ async def _prepare_run(
...
+    exec_dests = {input.dest for input in artifacts.inputs}
+    for src, dest in package.get_execution_files(code):
+        if dest not in exec_dests:
+            artifacts.inputs.append(GradingFileInput(src=src, dest=dest))
+            exec_dests.add(dest)
+    # Only interpreted languages (EXECUTION-kind scanner) contribute here; C++ and
+    # other compiled languages short-circuit inside expand() without walking.
+    dep_graph = deps_graph.expand(code, require_kind=DependencyKind.EXECUTION)
+    if dep_graph is not None:
+        # The entry script is symlinked from the content-addressed cache, so the
```
rsalesc: AFAIU, the problem happens because we always symlink the executable generated by the compilation step that is added to the cache.

can we instead just copy the file to the sandbox when the language is interpreted? there should be some bit in the langauge definition to identify it's interpreted, and we can use it. this generalizes better for newly added languages, not only for python

rbx/box/dependencies/graph.py
```diff
@@ -0,0 +1,61 @@
...
+    """Transitively discover ``code``'s quoted-include / relative-import dependencies.
+
+    Returns ``None`` when there is no scanner for the language, the source lives
+    outside the package root (remote/temporary files stay flat), or ``require_kind``
+    is given and the language's scanner does not contribute that kind. The last case
+    short-circuits **before** the (potentially expensive) transitive walk, so e.g. a
+    C++ source skips scanning entirely on the execution path. Cycle-safe; only
+    references that resolve to an existing file under the package root are followed.
+    """
+    # Lazy import avoids a code <-> dependencies import cycle.
```
rsalesc, line 38: i don't like the fact we're depending on the precise language name to decide which scanners should be applied.

notice in the linter case, the scanner is applied through specific linter registration in env.rbx.yml.

we could do the same for scanners (actually, I believe we *should add* the scanners field in env.rbx.yml just for completeness, and refer to scanners by their names in there), but I do believe an additional, more automatic approach is necessary.

I think we should formally introduce a concept called "language kind", where we default some builtin language kinds in rbx. we do have some functions such as "is_cxx", "is_cpp", "is_c", "is_python", "is_java", etc, that depend not only on the code item object, but usually on the actual command being executed.

one code can have more than one language kind (for instance cxx and cpp, or cxx and c). let's create this formal definition, and also apply scanners by it. let's take the opportunity to refactor current usages of "is_<kind>..." function with this new utility, it'd be clean.

rbx/box/code.py
rsalesc: Done in 30df78f. The interpreted entry script is now copied into the sandbox instead of symlinked from the cache (`GradingFileInput.symlink=False`), so its realpath stays inside the sandbox and `sys.path[0]` resolves to the mirrored source dir — siblings resolve naturally and the PYTHONPATH workaround is gone. "Interpreted" is derived from the language definition via `environment.is_interpreted` (passthrough / no compile commands — the same signal `compile_item` uses), so it generalizes to any interpreted language, not just Python.

rbx/box/dependencies/graph.py
rsalesc: Done in 9029ee4. Added `rbx/grading/language_kind.py` with a formal `LanguageKind` enum + `command_kinds(cmd)` — a command/language can carry multiple kinds (`g++` → `{CPP, CXX}`, `gcc` → `{C, CXX}`, `javac`/`java` → `{JAVA, JVM}`, `kotlinc` → `{KOTLIN, JVM}`, `python` → `{PYTHON}`). The `is_*_command` predicates are now thin aliases over `command_kinds` (behavior-identical); deprecating them in favor of direct `command_kinds` use is tracked in #529.

`environment.language_kinds` derives a language's kinds from its toolchain (compile commands, else execution command), not the name string. Scanners now declare the `LanguageKind`s they apply to plus a registry name, and `graph.expand` dispatches by kind intersection — plus an explicit per-language `scanners` env field (mirroring `linters`), with the bundled env declaring defaults.

PR #536, opened 2026-06-08
feat(presets): registry of presets with picker (#535)

### LINKED ISSUE #535: Create a registry of presets
`rbx` should keep a registry of presets in the presets folder.

This registry should be use to feed a list that is used in all commands where the user must choose a preset (and it can't be inferred from an existing .local.rbx folder) such as when creating a contest, creating a problem, etc.

Instead of simply assuming the default preset should be used, we should take the presets from the registry, and show them to the user. Every preset must have a description (the description should be part of the preset itself, not of the registry).

The registry must accept all sorts of fetch infos, and thus it might be fetched from a remote url, another git repo, a local rbx folder (as is the case with the default preset), etc.

rsalesc:
## Summary

Implements a **registry of presets** (#535). Instead of silently assuming the
bundled `default` preset, commands that need a preset (and can't infer one from
a `.local.rbx` folder) now pick from a registry of known presets, each with a
description.

- **`Preset.description`** — new field (canonical, author-defined); the bundled
  default preset and the built-in registry carry one.
- **Merged registry** (`rbx/box/presets/registry.py`, `registry_schema.py`) — a
  built-in `rbx/resources/presets/registry.yml` (shipped with rbx) unioned with
  a user-local `<app-dir>/presets/registry.yml` (initially empty). User entries
  win on a name collision.
- **Picker, not silent default** — `get_preset_fetch_info_with_fallback`: an
  explicit `--preset` wins; else an active `.local.rbx` preset; else, in an
  interactive terminal, a `questionary` picker over the registry; else (no TTY)
  a clear error asking for `--preset`.
- **`rbx presets registry ls | add | rm`** — manage the user registry. `add`
  reads only the preset's metadata (direct read for local/bundled, shallow
  clone for remotes — no install-time prompts). Using a new remote `-p <uri>`
  interactively offers to register it.
- **Docs + schema** — presets page documents the registry + `description`; the
  `PresetRegistry` JSON schema is published via `dump_schemas.py`.

### Behavior change

The silent `default` fallback is gone. In a **non-interactive** context (CI /
scripts / no TTY) with no `-p` and no active preset, rbx now **errors** asking
for `--preset` (e.g. `--preset default`) instead of assuming `default`. This was
an explicit design decision; `default` remains trivially selectable.

Design & plan: `docs/plans/2026-06-08-preset-registry-design.md`,
`docs/plans/2026-06-08-preset-registry.md`.

Closes #535.

## Test Plan

- [x] New unit/CLI tests for the registry schema, merge precedence, picker,
      resolver (interactive picker vs non-interactive error), and
      `registry ls/add/rm` + auto-offer (`tests/rbx/box/presets/`).
- [x] Full presets suite green (181 incl. linters dir; no test-name collision).
- [x] `ruff check`/`format` clean repo-wide.
- [x] CLI smoke: `rbx presets registry ls` renders; `registry add default`
      reads the bundled preset with no network.
- [x] Broad suite: remaining failures verified **identical on the base commit**
      (pre-existing local `testlib.h`/sandbox/docker), so no regressions.
- [ ] Reviewer: confirm the non-interactive error behavior change is acceptable
      for any downstream scripts/CI that relied on the silent default.


rbx/box/presets/registry.py
```diff
@@ -0,0 +1,119 @@
...
+def find_in_registry(uri_or_name: str) -> Optional[RegistryPreset]:
+    for p in get_merged_registry().presets:
+        if p.name == uri_or_name or p.uri == uri_or_name:
+            return p
+    return None
+
+
+def is_interactive() -> bool:
+    return sys.stdin.isatty() and sys.stdout.isatty()
+
```
rsalesc: please move this to utils.py, and call it is_interactive_tty to not me ambiguous with communication/interactive problems.

rsalesc: Done in 02b48ae. Moved it to `rbx/utils.py` as `is_interactive_tty()` (you're right — `is_interactive` was already taken by the COMMUNICATION/interactive-problem concept in `summary.py` and the UI screens). Updated the resolver/offer callers to `utils.is_interactive_tty()` and the tests to monkeypatch it there. `registry.py` no longer needs `import sys`. Preset tests green.

PR #542, opened 2026-06-08
refactor(cache): rename .box cache directory to .rbx

### LINKED ISSUE #306: Rename .box folder to .rbx
It doesnt make sense to use the name .box anymore

rsalesc:
Closes #306.

The `.box` cache/build directory name no longer reflects the tool (`rbx`). This renames it to `.rbx`.

## Changes
- Add `CACHE_DIR_NAME = '.rbx'` and `LEGACY_CACHE_DIR_NAME = '.box'` constants in `rbx/config.py`; the name was previously hardcoded as `'.box'` across several modules.
- Point the per-problem cache (`package.py`), global precompilation cache (`global_package.py`), and warning-stack cache (`sanitizers/warning_stack.py`) at `CACHE_DIR_NAME`.
- `presets`: `clean_copied_package_dir` and the package-glob ignore default now strip/ignore **both** `.rbx` and the legacy `.box`, so a stale cache left over after an upgrade never leaks into a generated preset/package.
- gitignore templates for new packages now emit `.rbx/`; dev/testdata gitignores keep both names.
- Updated test fixtures, e2e assertions/runner/matcher, and the e2e README.

## Design decisions
- **No migration / no fallback read:** old `.box` caches are gitignored and regenerated, so they simply orphan and rebuild as `.rbx` on the next run.
- **Constant, not scattered literals:** prevents the "name duplicated everywhere" problem from recurring.

## Testing
- TDD: added `tests/rbx/box/test_cache_dirs.py` and a legacy-`.box`-strip assertion (red → green).
- Diffed full-suite failing node-IDs **with vs. without** these changes: **zero** new failures introduced. The remaining failures on my machine are pre-existing C++ compilation / sandbox / docker issues unrelated to this change.
- `ruff check` + `ruff format` clean; commitizen check passed.

### DIFF
--- a/rbx/config.py
+++ b/rbx/config.py
@@ -19,6 +19,12 @@ app = typer.Typer(no_args_is_help=True)
 _RESOURCES_PKG = 'rbx.resources'
 _CONFIG_FILE_NAME = 'default_config.json'
 
+# Name of the per-problem and global rbx cache/build directory.
+CACHE_DIR_NAME = '.rbx'
+# Pre-#306 cache directory name. New code never writes here, but it is still
+# stripped when packaging/seeding so stale caches don't leak into outputs.
+LEGACY_CACHE_DIR_NAME = '.box'
+
 
 def format_vars(template: str, **kwargs) -> str:
     res = template
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # rbx
+.rbx/
 .box/
 
 .DS_Store
--- a/rbx/box/global_package.py
+++ b/rbx/box/global_package.py
@@ -4,7 +4,7 @@ import pathlib
 import shutil
 from typing import Iterator, Type
 
-from rbx.config import get_app_path
+from rbx.config import CACHE_DIR_NAME, get_app_path
 from rbx.grading.caching import DependencyCache
 from rbx.grading.judge.cacher import FileCacher
 from rbx.grading.judge.sandbox import SandboxBase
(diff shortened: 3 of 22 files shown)


rbx/config.py
```diff
@@ -19,6 +19,12 @@
 _RESOURCES_PKG = 'rbx.resources'
 _CONFIG_FILE_NAME = 'default_config.json'
 
+# Name of the per-problem and global rbx cache/build directory.
```
rsalesc, line 22: there are still a bunch of places in the codebase where ".rbx" is used directly (Python files). can we use this?

rsalesc: Done in 8a8a77f — routed the test fixtures/helpers (conftests, e2e assertions/runner/matcher, the presets cleanup tests, and `test_cache_dirs`) through `CACHE_DIR_NAME` / `LEGACY_CACHE_DIR_NAME` instead of the literal `.rbx`/`.box`.

After the rename there were no remaining direct cache-dir literals left in non-test source. I left these as-is on purpose, since they're a different concept from the cache directory (using `CACHE_DIR_NAME` for them would be wrong):
- `.local.rbx` — the installed-preset directory
- `*.rbx.yml` (problem/contest/env/preset) and the statement `.rbx` file-name suffix

If you meant the `.local.rbx` occurrences (hardcoded in ~12 spots) and want a dedicated constant for those too, I can do that in a follow-up — let me know.

Issue #326, opened 2025-10-29
Add quality of life improvements to `rbx ui`


rsalesc:
1. Show which solution was used as main
2. Show the actual testplan line that was used along the file line itself
3. Add filter to show only failing tests
4. Add a UI to navigate across tests first, and show verdicts of each solution (inverted UI)
5. Add filter to jump to test (fuzzy, use the test number, the generator and the testplan line in the search)
6. Add goto command to jump to test
7. Add faster navigation (jump multiple)

rsalesc:
Fanned these 7 ideas out into 4 themed child issues (high-level design in `docs/plans/2026-06-08-rbx-ui-qol-design.md`):

- **#547** — Display polish: show the main solution + inline testplan line content _(ideas 1, 2)_
- **#548** — Test-list filtering + fancy search box: failing-only filter, fuzzy search, goto _(ideas 3, 5, 6)_
- **#549** — Faster navigation: half-page, page, vim count prefixes _(idea 7)_
- **#550** — Inverted test-first explorer: navigate by test, show every solution's verdict _(idea 4)_

Suggested order: **#547 → #548 → #549** (#548 introduces a reusable test-list filter/index model that #549 builds on); **#550** is largely standalone. Goto (idea 6) was folded into #548's search box rather than a separate numeric command.

PR #551, opened 2026-06-08
feat(ui): show main solution badge in run-explorer

### LINKED ISSUE #547: rbx ui: show main solution + inline testplan line content
Part of #326 (umbrella). **Effort:** Small (~½ day) · **Risk:** Low · **Depends on:** nothing.

Two small, independent display enhancements to the `rbx ui` run-explorer flow. No new screens.

## #1 — Surface the MAIN solution

The "main" solution is the first solution with `outcome: accepted` (`package.get_main_solution()`, `rbx/box/package.py:462`). A `MAIN` badge is **already** rendered in `RunScreen`/`SolutionReportScreen` by `_build_solution_selection_label` (`rbx/box/ui/screens/run.py:27`) but is **not** surfaced in the run-explorer flow.

- Extract the badge-styling logic into a shared helper (e.g. `run_ui.get_main_badge(solution)` → markup or empty string).
- `RunExplorerScreen`: have `get_solution_markup` (`rbx/box/ui/utils/run_ui.py:161`) append the badge so the solution list marks the main solution.
- `RunTestExplorerScreen`: prefix `self.title` (`rbx/box/ui/screens/run_test_explorer.py:77`) with a `[MAIN]` marker when `self.solution.path == get_main_solution().path`.

## #2 — Inline testplan line content

Today `get_generation_metadata_markup` (`rbx/box/testcase_extractors.py:132`) renders `metadata.generator_script` as just `path:line` (the `__str__` of `GeneratorScriptEntry`, `rbx/box/generation_schema.py:18`). The *content* of that line is never read.

- When `metadata.generator_script` is set, read line `entry.line` from `entry.path` and render it alongside the location, e.g. `Gen. script: gen.txt:42 → gen_random 1000 50`.
- **Edge cases:**
  - File edited since build → line content may be stale: render anyway from the current file (a small dim caveat is acceptable; do not error).
  - `@input { … }` block syntax spans multiple lines — `line` points at the start; show the first line + `…`.
  - File missing / line out of range → fall back to today's `path:line`-only rendering.
  - Read once per entry and cache (avoid re-opening the file on every footer repaint).

## Tests

- `get_main_badge` / `get_solution_markup` marks the main solution and no other.
- Metadata markup renders line content; graceful fallback when the file/line is missing.

---
_Design: `docs/plans/2026-06-08-rbx-ui-qol-design.md`._

rsalesc:
Part of #326. Closes #547.

Surfaces the MAIN solution across the `rbx ui` run-explorer flow. No new screens.

## Surface the MAIN solution
- Extracted the badge-styling logic into shared helpers `run_ui.is_main_solution` / `run_ui.get_main_badge` (markup or empty string).
- `RunExplorerScreen`: `get_solution_markup` now appends the badge, so the solution list marks the main solution.
- `RunTestExplorerScreen`: `self.title` is prefixed with a `[MAIN]` marker when the solution is the main one.
- Rewired `_build_solution_selection_label` (RunScreen/SolutionReportScreen) to reuse `get_main_badge`, so the badge is consistent across screens.

## Inline testplan line content — dropped
The issue's second item (inline the generator-script line content) was implemented and then reverted after review: the parser sets `generator_call`/`copied_from` alongside `generator_script`, so the existing `Gen. call:`/`Copied from:` lines already show that content for the common cases. Only inline `@input` literals would gain anything, which didn't justify the file-read + caching machinery. See commit 32b4d58.

## Tests
- `is_main_solution` / `get_main_badge` / `get_solution_markup` mark the main solution and no other; `_build_solution_selection_label` shows the badge for main and the outcome otherwise.
- `RunTestExplorerScreen` title gets the `[MAIN]` prefix only for the main solution.

All new/changed tests pass; ruff check/format clean. (Pre-existing C++/sandbox compile failures in CI — `testlib.h: No such file or directory`, generator compile errors — are environmental and unrelated to this display-only change.)

### DIFF
--- a/rbx/box/testcase_extractors.py
+++ b/rbx/box/testcase_extractors.py
@@ -1,4 +1,5 @@
 import abc
+import functools
 import pathlib
 from typing import Iterable, List, Optional, Set
 
@@ -129,6 +130,36 @@ def get_testcase_metadata_markup(entry: GenerationTestcaseEntry) -> str:
     return '\n'.join(lines)
 
 
+@functools.lru_cache(maxsize=4096)
+def _read_generator_script_line(path: pathlib.Path, line: int) -> Optional[str]:
+    """Read the 1-based ``line`` from a generator script, for inline display.
+
+    Returns the stripped line content (suffixed with ``…`` when the line opens
+    a multi-line ``@input { … }`` block), or ``None`` when the file is missing
+    or the line is out of range -- callers then fall back to a location-only
+    rendering.
+
+    The result is cached so repeatedly repainting the run-explorer footer does
+    not re-open the file. The content reflects the file as it is *now*, which
+    may differ from what was generated at build time; that staleness is
+    acceptable.
+    """
+    try:
+        file_lines = path.read_text().splitlines()
+    except (OSError, UnicodeDecodeError):
+        return None
+    if line < 1 or line > len(file_lines):
+        return None
+    content = file_lines[line - 1].strip()
+    if not content:
+        return None
+    if content.endswith('{'):
+        # ``@input { … }`` block opener: the directive spans multiple lines and
+        # ``line`` points at its start, so flag that more content follows.
+        content = f'{content} …'
+    return content
+
+
 def get_generation_metadata_markup(metadata: GenerationMetadata) -> str:
     lines = []
     if metadata.copied_from is not None:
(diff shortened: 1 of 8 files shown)


rbx/box/testcase_extractors.py
```diff
@@ -140,9 +171,12 @@ def get_generation_metadata_markup(metadata: GenerationMetadata) -> str:
             f'[bstatus]Gen. call:[/bstatus] {utils.escape_markup(str(metadata.generator_call))}'
         )
     if metadata.generator_script is not None:
-        lines.append(
-            f'[bstatus]Gen. script:[/bstatus] {utils.escape_markup(str(metadata.generator_script))}'
-        )
+        entry = metadata.generator_script
```
rsalesc: I suspect this change isn't really necessary since usually the generator call field above would be filled already with the content of the line (or at least something very close to it). Could you check?

rsalesc: Checked — you're right for the cases that matter. The parser sets `generator_call`/`copied_from` on the same `ScriptGeneratedInput` as `generator_script`, and `str(GeneratorCall)` is literally `gen_random 1000 50`, so the `Gen. call:`/`Copied from:` lines already show the line content for generator calls and `@copy`. The only case that would gain anything is inline `@input` literals (whose `content` isn't rendered today), and that doesn't justify the file-read + caching machinery. Dropped the whole inline change in 32b4d58; the MAIN-solution badge work stays.

PR #552, opened 2026-06-08
refactor(timing): collapse get_solution_limits family (#351)

### LINKED ISSUE #351: The family of functions called `get_solutions_limit` is a bit weird, rethink

rsalesc:
## What

Closes #351 — the `get_solution_limits` family was confusing and bug-prone. Two overlapping skeleton resolvers plus a `time is None` disk-reload dance.

### Root cause

`Limits.time` was overloaded: it meant **both** the *declared* time limit and the *enforced* one. `get_limits_for_language` nulls `time` to signal "do not enforce a TL for this run", which destroyed the declared value that the timing report wants to display. So `_print_timing` had to re-resolve the profile from disk (`get_solution_limits_from_disk`) whenever `time` was `None`:

```python
limits = skeleton.get_solution_limits(solution)
if limits.time is None:
    limits = skeleton.get_solution_limits_from_disk(solution)  # reload to recover the TL
assert limits.time is not None
```

### Change

Separate the two meanings on the `Limits` model:

- **`Limits.configuredTime`** — the declared TL, preserved through enforcement nulling.
- **`Limits.display_time()`** — returns `configuredTime` (falling back to `time`) for reporting.
- **`Limits.time`** keeps its exact meaning: the *enforced* limit (`None` = no enforcement).

The timing report's no-eval branch now reads `display_time()` directly, so **`get_solution_limits_from_disk` is deleted** and the disk-reload is gone. A single resolver (`get_solution_limits`) remains.

### Behavior preservation

- Sandbox (`_get_execution_config`), checkers / soft-TLE / double-TL, and packaging all still read `limits.time` — unchanged.
- The run cache key uses sandbox params, not the `Limits` object, so the new field is cache-safe; old cached `.eval`/`.log` logs deserialize via the field default (and `display_time()` falls back to `time`).
- `limits_info.*` names left untouched (deliberate scope cut; the broader rename is the remaining half of "rethink the family").

Design doc: `docs/plans/2026-06-08-limits-rethink-design.md`.

## Tests

- `tests/rbx/grading/limits_test.py` — `display_time()` fallback / preference / none.
- `tests/rbx/box/tasks_test.py::TestGetLimitsForLanguage` — `configuredTime` survives `use_timelimit=False` nulling; `timelimit_override` sets both.
- `tests/rbx/box/limits_info_test.py` — `get_limits` records `configuredTime`.
- `tests/rbx/box/solutions_test.py` — skeleton resolver recovers the declared TL via `display_time()` with no disk reload.

All affected suites green locally (pre-existing `testlib.h not found` C++ env failures on this machine are unrelated).

### DIFF
--- a/docs/plans/2026-06-08-limits-rethink-design.md
+++ b/docs/plans/2026-06-08-limits-rethink-design.md
@@ -1,0 +1,122 @@
+# Rethinking the `get_solution_limits` family (#351)
+
+## Problem
+
+Issue #351 (`bug`, `timing`) flags that the cluster of "get the limits" helpers is
+confusing and bug-prone. The concrete smells:
+
+1. **Two overlapping skeleton methods** on `SolutionReportSkeleton`
+   (`rbx/box/solutions.py`):
+   - `get_solution_limits(solution)` — looks up a **pre-computed, cached**
+     `Limits` from `self.limits` (keyed by language), built once at skeleton
+     creation via `get_limits_for_language`.
+   - `get_solution_limits_from_disk(solution)` — **re-reads the profile YAML
+     from disk** and recomputes a fresh `Limits`, bootstrapping off the cached
+     variant just to learn the profile name.
+
+2. **The disk-reload dance** at the timing-report call site:
+
+   ```python
+   limits = skeleton.get_solution_limits(solution)
+   if limits.time is None:                                # cached time was stripped...
+       limits = skeleton.get_solution_limits_from_disk(solution)  # ...reload to recover it
+   assert limits.time is not None
+   ```
+
+   `get_limits_for_language` (`tasks.py`) deliberately nulls `Limits.time` to
+   signal "don't enforce a TL for this run" (`use_timelimit=False`, or
+   `time <= 0`). That destroys the *declared* time limit that the reporting code
+   wants to display, forcing a round-trip back to disk.
+
+3. **Name confusion** in `limits_info.*` (out of scope here, but the reason the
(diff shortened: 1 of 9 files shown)


rsalesc:
## Adversarial review pass

Spawned an adversarial reviewer to attack the change. It found **one real regression** I'd introduced, now fixed in `0d61bc8`:

**`rbx time` estimation printed `Time limit: -1 ms`.** Time-limit estimation runs solutions with `timelimit_override=-1` (the "unlimited" sentinel, `timing.py:377`). I'd copied the override into `Limits.configuredTime`, and since the timing report's no-eval branch reads `display_time()`, it surfaced the raw `-1`. The old `get_solution_limits_from_disk` re-resolved the profile via `get_limits(...)` and *silently ignored the override*, so it showed the real declared TL — my "equivalence" claim missed that.

Fix: `timelimit_override` now affects only the enforced `time`; `configuredTime` stays the declared profile TL. Regression test added (`test_unlimited_override_does_not_leak_into_displayed_limit`), and `test_timelimit_override_...` updated to assert the override no longer overwrites the declared limit.

Everything else held up under attack (verified clean): `time` semantics unchanged for sandbox/checkers/double-TL/packaging; cache key unaffected (uses sandbox params, not `Limits`); old cached logs deserialize via the field default; `get_solution_limits_from_disk` has no remaining references; no aliasing leaks. One cosmetic NIT (the `console.print(limits)` in stress mode now shows `configuredTime`) was considered and kept — it's accurate, informative info in find-slowest mode.

PR #573, opened 2026-06-09
perf(completion): fast static-spec shell completion (#333)

### LINKED ISSUE #333: <tab> autocomplete is slow in linux
Takes about 1s in my machine

rsalesc:
## Summary

Fixes #333 — `<tab>` shell completion was ~1s on Linux (~545–741ms measured locally) because every completion ran the `rbx` binary, which imported the **entire CLI** (`textual`, `mechanize`, `bs4`, `lxml`, GitPython, grading engine, pydantic schemas…) before computing a single completion.

This serves completions from a **precomputed static spec generated from Typer**, intercepting in `main.py` *before any heavy import*. Result: **~545–741ms → ~76–80ms (~7–9× faster)**.

### How it works
- A generator (`generate.py`) introspects the live Typer/Click tree and serializes a committed static spec (`_spec.py`). It ships in the wheel, so it can't be stale at runtime; a **drift test** regenerates and diffs it in CI.
- `main.py:app()` calls `completion.entry.handle_completion()` first; on a completion request it resolves against the static spec and reuses Click's **native** per-shell formatters for byte-compatible output. The heavy CLI is never imported.
- **No slow path ever:** unknown/path positions and any error hand off to the shell's own file completion (`file` directive).

### Extensibility (hooks for new completers)
A small completer registry. Adding a completer is a 3-step recipe (documented in `rbx/box/completion/CLAUDE.md`):
1. write `complete_x(ctx, incomplete)` in `completers.py`,
2. `@register_completer('x')`,
3. attach `autocompletion=annotations._adapt('x')` and run `mise run gen-completion-spec`.

Demonstrated live by wiring the `language` completer onto the real `--language` options (e.g. `rbx tool convert --language <tab>` → `c++17`, `exinc`). Note: those completers were previously vestigial/unwired; the registry makes wiring the rest trivial.

## Guarantees (all in CI)
- **740-case differential test** — engine output equals real Typer completion across the whole command tree (incl. aliases, `--opt=value`, supplied-option suppression, hidden commands, auto `--help`).
- **Import-firewall test** — fails if any heavy module rejoins the completion path.
- **Drift test** — committed spec must match regeneration.
- Robustness tests (the fast path never crashes the shell) + a `mise run bench-completion` benchmark.

## Test plan
- [x] `uv run pytest tests/rbx/box/completion -q` → 793 passed
- [x] Full non-slow suite → 3285 passed; the 33 failures are pre-existing/environmental (g++ unavailable on this machine), unrelated to this change
- [x] `uv run mise run bench-completion` → ~76–80ms (was ~545–741ms)
- [ ] Reviewer: verify `<tab>` behaves correctly in a real bash/zsh session

Design + plan: `docs/plans/2026-06-09-fast-completion*.md`.


rbx/annotations.py
```diff
@@ -18,39 +13,36 @@ class _PackagePathMarker:
 PackagePath = _PackagePathMarker()
 
 
-def _get_language_options():
-    return sorted(get_config().languages.keys())
-
+def _adapt(key: str):
```
rsalesc, line 16: can you give a better name to this? also, it seems this should not be a private function

Issue #570, opened 2026-06-09
statements v2: docs update


rsalesc:
Part of #556. See `docs/plans/2026-06-09-statements-v2-design.md` §3-§6.

**Depends on:** #565, #566

## Scope
Update the statements docs: `rbx/box/statements/CLAUDE.md` and the docs-site pages for the v2 model — schema, namespaces, overlay/path-resolution, `extends`.

## Acceptance
Docs build (non-strict) reflects v2; inner CLAUDE.md updated.

<sub>Tracking code: S14</sub>

Issue #569, opened 2026-06-09
statements v2: tests + e2e port


rsalesc:
Part of #556. See `docs/plans/2026-06-09-statements-v2-design.md` §7.

**Depends on:** #565, #566

## Scope
Port unit tests and `tests/e2e/*` fixtures to the v2 schema + overlay. Remove tests for deleted concepts (`override`/`inheritOverride`/`inheritFromContest`/`configure`/`steps`).

## Acceptance
Suite green on v2; e2e statement scenarios pass.

<sub>Tracking code: S13</sub>

Issue #568, opened 2026-06-09
statements v2: Polygon packager update


rsalesc:
Part of #556. See `docs/plans/2026-06-09-statements-v2-design.md` §2/§6.5.

**Depends on:** #566

## Scope
Consume the now-portable, self-contained TeX; remove the absolute-path special-casing; flip `externalize`/`demacro` at export time (they are no longer schema fields).

## Acceptance
Polygon export produces a valid statement package from the v2 overlay; externalize/demacro forced by the packager; tests.

<sub>Tracking code: S12</sub>

Issue #567, opened 2026-06-09
statements v2: preset templates + bundled defaults


rsalesc:
Part of #556. See `docs/plans/2026-06-09-statements-v2-design.md` §7.

**Depends on:** #564, #565, #566

## Scope
Rewrite the bundled preset templates (problem template, contest template, sample-explanation templates, infosheet) for v2: `\subimport` joins, `params`/`vars`/`contest.vars` namespaces, per-sample folders, root-relative I/O. Update the default preset `problem.rbx.yml` / `contest.rbx.yml`.

## Acceptance
`rbx create` default preset builds problem + contest statements under v2; goldens updated.

<sub>Tracking code: S11</sub>

PR #574, opened 2026-06-09
statements v2: phase 2 + 3 — engine core, standalone & contest builds (#560-#566)

rsalesc:
Implements **Phase 2 (engine core)** and **Phase 3 (builders & build modes)** of the statements v2 plan ([#556](https://github.com/rsalesc/rbx/issues/556)), per `docs/plans/2026-06-09-statements-v2-design.md`. Builds on the schema/`extends` work from #572. The builders stay similar (block extraction, the LaTeX Jinja env, and the `tex→pdf` pdflatex loop are reused); what changed is the configuration surface and the path-resolution model.

## Phase 2 — engine core (new modules)
- **`resolver.py` (S4→#563):** contest-required hard error; standalone candidate resolution (`0`/`>1` are distinct config errors); join matcher by `(language, variant)` + matching rbx type.
- **`overlay.py` (#560):** the stager — merged standalone overlay with name-collision detection, isolated `.problems/<SHORT>/` join layout, contest-chrome overlay. Never parses/rewrites TeX, only places files.
- **`context.py` (#561):** namespaced template context — `params` (statement) / `vars` (problem or contest) / `contest.vars` kept **distinct, unmerged**, plus `problem.import_dir`/`import_file` and the `sample.*` handles.
- **`sample_staging.py` (#562):** per-sample `.samples/<idx>/` folders — `in`/`out` mirrored (root-relative for `\VerbatimInput`), explanation rendered to `explanation.tex` with its source dir overlaid for figures (base-relative for `\subimport`), interactive chunks.

## Phase 3 — builders & build modes
- **`render.py` + `engine.py` (#564):** full-document vs fragment render modes sharing the block-extraction core; pdflatex runs from the overlay **root** so the generated TeX is portable.
- **`build_statements.py` (#565):** `rbx st b` end-to-end — resolve contest → stage merged overlay → render `standaloneProblemTemplate` → compile → `build/statement-<lang>[-<variant>].pdf`.
- **`contest/build_contest_statements.py` (#566):** `rbx contest st b` — per-problem fragments via `contestProblemTemplate` joined through `\subimport`, plus `documents` emitted without joining. CLI wired to also build documents; fixed the packaging contest-build call site to pass `output_type` by keyword.

## Path-resolution contract (proved by the #557 spike)
- `\subimport` / `\includegraphics` / `\input` are import-base-relative (`import.sty` rebases at every depth — no `\graphicspath`).
- `\VerbatimInput` (sample I/O) does **not** honor the import base → I/O handles are root-relative; explanations go through `\subimport` (base-relative). Both relative → portable overlay.

## Out of scope (deferred per plan)
Phase 4: preset template rewrite (#567), Polygon packager update (#568 — its build path stays stubbed and keeps importing `builders.py`'s v1 classes), full test/e2e port (#569), docs site (#570). `tutorials` build commands and the contest-less default-template fallback remain deferred.

## Test plan
- 42 new unit + integration tests (resolver / overlay / context / sample staging / render / standalone build / contest join). Full statements+contest+packaging (non-docker) suite: **507 passed**.
- Both paths verified against **real `pdflatex`**: standalone overlay (`\example` root-relative I/O + `\subimport` explanation + figure via `import.sty`) and the contest join (two problems via `\subimport`) compile to valid PDFs.
- `ruff check`/`format` clean; `rbx --help`, `rbx st --help`, `rbx contest st --help` load.
- Full default suite: 47 pre-existing failures only (C++/checker/validator/sandbox/walltime/docker), confirmed by re-running the suspicious subset on the stashed baseline — zero new failures.

### DIFF
--- a/rbx/box/statements/build_statements.py
+++ b/rbx/box/statements/build_statements.py
@@ -40,6 +40,8 @@ _V2_POLYGON_PENDING = (
 
 
 def get_environment_languages_for_statement() -> List[StatementCodeLanguage]:
+    """The configured environment languages exposed to templates as ``languages``
+    (id, readable name, and the compile/run command shown in limit tables)."""
     env = environment.get_environment()
 
     res = []
@@ -69,6 +71,8 @@ def get_environment_languages_for_statement() -> List[StatementCodeLanguage]:
 def get_builder(
     name: ConversionType, builder_list: List[StatementBuilder]
 ) -> StatementBuilder:
+    """Look up the (legacy v1) builder for a conversion type, or exit. Kept for
+    the deferred Polygon export path; the v2 build uses ``render.py`` directly."""
     candidates = [builder for builder in builder_list if builder.name() == name]
     if not candidates:
         console.console.print(
@@ -81,6 +85,14 @@ def get_builder(
 def get_implicit_builders(
     input_type: StatementType, output_type: StatementType
 ) -> Optional[List[StatementBuilder]]:
+    """Shortest chain of (legacy v1) builders converting ``input_type`` to
+    ``output_type``, or None if unreachable.
+
+    A BFS over ``BUILDER_LIST`` treating each builder as an edge
+    ``input_type -> output_type``; ``par`` maps a reached type to the builder
+    that produced it, then the chain is walked back from ``output_type``. Legacy
+    v1 machinery kept for the deferred Polygon path.
+    """
     par: Dict[StatementType, Optional[StatementBuilder]] = {input_type: None}
 
     def _iterate() -> bool:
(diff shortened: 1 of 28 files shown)


rbx/box/statements/build_statements.py
```diff
@@ -234,17 +240,74 @@ async def build_statement_bytes(
...
 
 def needs_samples(statement: Statement) -> bool:
-    raise NotImplementedError(_V2_PENDING)
+    return statement.samples
+
+
+def _overlay_dir(statement: Statement) -> pathlib.Path:
+    root = (
+        package.get_statements_build_path()
+        / 'st'
```
rsalesc, line 323: Why this? Doesn't this already come from `get_statements_build_path()`? if not, can we put this behind a function that explains what this is for?

rsalesc: Good call — clarified in c7b1585e.

`get_statements_build_path()` is the shared `build/statements` dir (it also holds e.g. interactive-sample chunk folders), so it is *not* per-statement. Each standalone build needs its own scratch overlay tree (the merged contest-chrome + problem-dir overlay that pdflatex runs from), so I key it by `(language, variant)` to avoid one statement clobbering another. I renamed `_overlay_dir` → `_standalone_overlay_root` with a docstring spelling this out, and documented `get_statement_build_path` to make clear it is the *final* artifact at the package build root (`build/statement-<lang>.pdf`), distinct from this scratch dir.

Issue #556, opened 2026-06-09
Design statements v2


rsalesc:
We plan to rework statements entirely in the v1 release of rbx.

Nowadays, there's a huge mess about statement inheritance, param overriding and such. Also, there's a huge mess on assets coming from contests, problems or sample explanations, which create a huge mess on cwd resolution.

We plan to simplify the statement engine (at least from a configuration/YAML level, the builders can still be pretty much similar) to make use simpler for users.

---

## Problem statement proposal

Problem-level statements should be declared in two different sections `statements` and `tutorials`.

They should be simple, and should have a type similar to today. By default, they should be `rbxTeX`.

```
statements:
   - language: en
      file: "statements/statement-en.rbx.tex"
      variant: "..." # optional string field to disambiguate between different formats of the same language
      type: rbxtex | rbxmd | tex | md | pdf
      params:
           key1: value1
           key2: value2
```

Notice there are no steps param, no configure params, just a barebones statement declaration. Also, there's no need to declare assets. Basically, we expect all assets in a package that are under a certain to-be-built statement file to automatically be included during the statement building phase in the temp directory, and thus all assets to be accessible.

Notice the pair (language, variant) must be unique.

The `tutorials` field should be pretty similar, except that it's in a different section.

And here's the biggest catch: problem-level statements can't be built unless they're part of a contest. I'll explain this further in the next section

## Contest statement proposal

Contest statements should be declared in two different sections `statements` and `tutorials` too.

They should be simple and should have a type similar to today. By default, they should be `rbxTeX`.

```
statements:
   - name: "statement-en"
      language: en
      file: "statements/contest-en.rbx.tex"
      variant: "..." # optional string field to refer to a problem-level statement it's associated with
      type: rbxtex | rbxmd | tex | md | pdf
      params:
           key1: value1
           key2: value2
      standaloneProblemTemplate: "statements/problem-template.rbx.tex"
      contestProblemTemplate: "statements/contest-template.rbx.tex"
```

Notice the fields variant, params, and the template fields only make sense for rbxtex and rbxmd statements, since they're the only ones that can join problems.

rbxtex and rbxmd statements expect the matching problem statements (matched by the pair (language, variant)) to have the corresponding type for the join.

Notice the pair (language, variant) in a contest-level statement DOES NOT need to be unique, only the name needs to be.

Again, `tutorials` should be very similar.

The question probably is: why there are three templates being specified in there (file, standaloneProblemTemplate, contestProblemTemplate)?

[...]

PR #577, opened 2026-06-10
statements v2: preset templates + bundled defaults (#567)

rsalesc:
Part of #556 (statements v2), **phase 4 — S11 / #567**. Ports the bundled `default` preset to the statements-v2 engine that landed in phases 0–3, so `rbx create` / `rbx contest` produce packages that actually build (the old v1 preset YAML no longer parses against the v2 schema).

## What changed

**Preset restructure — the contest owns the chrome + templates**
- New `contest/statements/` chrome dir (real files, no symlinks): `icpc.sty`, `logo.png`, `instructions.tex`, plus the v2 templates:
  - `contest.rbx.tex` — task-sheet joining doc (`\subimport`s each problem via the `import_dir`/`import_file` handles)
  - `problem.rbx.tex` (standalone full doc) + `problem-fragment.rbx.tex` (contest fragment), sharing a `_problem-body.rbx.tex` Jinja partial so a problem renders identically standalone and joined
  - `editorial.rbx.tex` + `editorial-fragment.rbx.tex` (EDITORIAL watermark, solutions)
  - `info.jinja.tex` — info sheet (per-problem limits table) as a `documents` entry
- `contest.rbx.yml`: two rbxTeX statements (`statement-en`, `editorial-en`) on `(en, default)` — only `statement-en` carries a `standaloneProblemTemplate` so `rbx st b` resolves unambiguously — plus an `info-en` **document**.
- `problem.rbx.yml`: v2 statement at `statement/statement.rbx.tex`; `show_limits` moves to the statement's `params` namespace; samples moved to `statement/samples/`.
- Dropped the `shared/` dir and the cross-package symlinks — the problem reaches the chrome through the build overlay, not copies.

**Namespace migration in templates**: `problem.vars.show_limits`→`params.show_limits`, `N.max`→`vars.N.max`, `sample.inputPath`→`sample.input` (root-relative, for `\VerbatimInput`) with `\subimport`-ed explanations, `problem.path|parent`→`problem.import_dir`.

**Engine: documents can read problem metadata**
The v1 info sheet iterates all problems for a limits table. Rather than model it as a joining rbx statement (its first cut here), the engine now passes a **metadata-only** `problems` list (title / short_name / limits / profiles / groups — no blocks, samples, or `\subimport` handles) to contest **documents** too. Documents still never import problem statements or samples; they can just read metadata. So the info sheet is a plain `jinja-tex` `documents` entry. Design §4 and the inner `CLAUDE.md` updated; documents now build under the active timing profile.

## Verification
- v2 schema parses; `expanded_statements` / `expanded_documents` / params resolve.
- **Real `pdflatex`** compiled standalone, contest task sheet, editorial (watermark), and the info document (limits table); content checked (`| sci`→`10^{9}`, `1000 ms` / `256 MiB`, root-relative sample I/O, `\subimport` explanations, per-problem table A=1000 / B=2000).
- **Real build glue** (`execute_build` standalone + contest, `build_document` with metadata) on the actual preset files.
- New unit test: documents receive per-problem metadata. Fixed the two tests that referenced the moved preset paths. Suites green: **statements + contest + presets + seed + creation** all pass. ruff clean.

## Follow-ups (not in this PR — scoped to #567 only)
- The `default-preset` e2e does standalone `rbx st b` on a contest-less problem, which v2 hard-errors (contest required; the fallback is the deferred S15). Rewiring it to seed a contest belongs to **#569**. e2e is excluded from default CI, so the unit suite stays green.
- Remaining phase 4: **#568** (Polygon), **#569** (tests/e2e port), **#570** (docs).


rbx/resources/presets/default/problem/statement/statement.rbx.tex
rsalesc: let's please have the editorial in a separate `editorial.rbx.tex` file in the same dir

rbx/resources/presets/default/problem/problem.rbx.yml
```diff
@@ -29,18 +29,15 @@ solutions:
...
     outcome: "INCORRECT"
 statements:
-  - name: "statement-en"
-    path: "documents/statement.rbx.tex" # Open this file to edit your statement.
+  # A problem statement is built inside a contest (statements v2): the contest
+  # owns the LaTeX templates, and `rbx st b` borrows the matching one. Match is
+  # by (language, variant); `variant` defaults to "default".
+  - language: "en"
+    file: "statement/statement.rbx.tex" # Open this file to edit your statement.
     type: "rbxTeX"
```
rsalesc, line 34: let's omit such things in the preset

rbx/resources/presets/default/problem/statement/statement.rbx.tex
rsalesc: Done in fa82befe — the editorial now lives in `problem/statement/editorial.rbx.tex`; the `editorial` block just `%- include`s it (kept it a block rather than a separate statement variant so `rbx st b` isn't forced to build it standalone). One wrinkle: in a standalone build the problem dir is merged with the contest chrome, so the contest's editorial joining doc had to be renamed `editorial-sheet.rbx.tex` to avoid colliding with the problem's `editorial.rbx.tex`.

rbx/resources/presets/default/problem/problem.rbx.yml
rsalesc: Done in fa82befe — trimmed the verbose explanatory comments across the preset down to short role comments. Kept `show_limits: true` (one concise inline comment): dropping it would silently hide limits since the param wrapper returns Undefined and `truthy(undefined)` is False. Say the word if you'd rather drop the param too and have the template default to showing limits when unset.

rbx/resources/presets/default/problem/statement/statement.rbx.tex
```diff
@@ -15,5 +15,5 @@
 %- endblock
 
 %- block editorial
```
rsalesc: I meant this whole block should be moved to a different file. The system should be able to support this... if it doesn't, let's brainstorm a way to do so.

Basically, the editorial.rbx.tex should be the "file" entry of a tutorial for the problem.

rsalesc: You were right — the system didn't support it. The v2 *schema* already had `tutorials` (problem + contest), but phases 2–3 never wired a build path or CLI for them. Implemented in d8237a81:

- Engine: threaded a `StatementKind` through the existing build entry points (selects `expanded_tutorials` vs `expanded_statements` on both problem and contest + the `tutorial-<lang>` output prefix; the resolver was already list-parameterised). New parallel `tutorials, tut` command app: `rbx tut b` (standalone) and `rbx contest tut b` (join), mirroring `rbx st b`. Contest `documents` build only under the statements command.
- Preset: the editorial is now a real **tutorial** — the problem has a `tutorials` entry whose `file` is `statement/editorial.rbx.tex` (its own rbxTeX file), and the contest has an `editorial-en` tutorial with standalone + in-contest templates. `statement.rbx.tex` no longer has an editorial block.

Verified with real pdflatex: `rbx tut b` → `build/tutorial-en.pdf`, `rbx contest tut b` → joined editorial PDF. Added fixture + standalone/contest tutorial tests (487 green). One default-format call: the editorial shows title + solution (it renders its own file, so it can't auto-pull the statement) — setters can expand `editorial.rbx.tex`.

PR #578, opened 2026-06-10
fix(completion): dedupe command/option aliases to one candidate

rsalesc:
## Problem

`rbx <tab>` and `rbx run -<tab>` list **every alias** of each command/option on its own line — `build b … package pkg …`, `--verification-level --verification -v …` — which is noisy and disorganized.

## Why not group them in the shell

The original attempt (earlier commits, now reverted) tried zsh `_describe` grouping. Two problems surfaced when testing against a real oh-my-zsh + fzf config:

1. A **custom directive type** (`grouped`) emitted by the binary breaks any *already-installed* completion function — it drops the unknown type and falls back to listing the whole directory (100+ bogus entries).
2. `_describe` itself **renders as a broken two-block layout** (all names, then all descriptions) under that config (fzf rebinds TAB; oh-my-zsh's `matcher-list` reorders), even though it renders fine in a vanilla shell.

## Fix: application-level dedup

Collapse aliases in the engine instead of the shell. `_first_match` returns the first name (in declaration order) matching the incomplete, and both subcommand- and option-name completion emit **one candidate per command/option**:

- broad prefix (``, `-`, `--`) → canonical full name (`package`, `--verification-level`)
- prefix typed toward a specific alias / `--no-` form → that exact spelling still completes (`pkg`, `-v`, `--no-check`)

The wire protocol and the zsh template are **unchanged** (`plain`/`file`/`dir` + `compadd -d`), so it's robust across shells, needs no new completion install, and degrades gracefully. Only NAME completion is deduped — option/positional **values** still match Typer exactly.

## Verification

- 841 completion tests pass. `differential_test` now asserts every offered name is a real Typer name and that exactly one is shown per command/option (values keep strict Typer parity).
- Rendered end-to-end in the reporter's real oh-my-zsh + fzf config: `rbx <tab>` → one clean inline line per command; `rbx run --<tab>` → one line per option, alphabetically sorted; aliases/`--no-` forms still complete when typed toward. No two-block, no directory-listing blowup.

### DIFF
--- a/rbx/box/completion/CLAUDE.md
+++ b/rbx/box/completion/CLAUDE.md
@@ -109,14 +109,20 @@ CI if the committed module is stale.
   `completion_init()` would change the output format and break byte-parity. (It
   also avoids global-registry pollution when tests import the heavy app.)
 - Command names are stored **raw**, comma-joined (`'package, pkg'`); the engine
-  splits on `', '` both for descent AND for completion. Subcommand completion
-  (`_command_name_items`) offers each name/alias as its OWN prefix-filtered
-  candidate (deduped in registration order), so a typed prefix completes to a
-  single concrete name (`pa` → `package`, `pkg` → `pkg`) instead of inserting the
-  unusable `'package, pkg'` string. This is a deliberate divergence from Typer
-  (which offers the raw joined string); `differential_test.py` checks command
-  names against a spec-derived expectation and keeps strict Typer-parity only for
-  option names and values.
+  splits on `', '` both for descent AND for completion. **Alias dedup:** both
+  subcommand-name completion (`_command_name_items`) and option-name completion
+  emit ONE candidate per command/option, via `_first_match` — the first name (in
+  declaration order) that matches the incomplete. A broad prefix (`''`, `-`,
+  `--`) yields the canonical full name (`package`, `--verification-level`); a
+  prefix typed toward a specific alias or `--no-` form completes that exact
+  spelling (`pkg`, `-v`, `--no-check`). This keeps `rbx <tab>` / `rbx run -<tab>`
+  from listing every alias on its own line — important because zsh alias
+  *grouping* (`_describe`) renders badly under some user configs (fzf TAB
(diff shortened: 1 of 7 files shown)


rsalesc:
## Follow-up: fixed a regression in the first commit

The first commit introduced a **new `grouped` zsh directive type** emitted by the binary. This broke real shells: any completion function already installed on a user's machine (via `rbx --install-completion`, or from an older rbx) has no branch for `grouped`, so it **silently drops every name candidate** and zsh falls back to listing the entire working directory — 100+ bogus entries for both `rbx <tab>` and `rbx run -<tab>`. (My isolated `zsh -f` test missed it; it only reproduces with an oh-my-zsh-style config + a pre-existing installed completion.)

**Deterministic proof of the mechanism** (the old #576 template's parse loop against each binary output):
```
A user with the OLD completion installed, completing 'rbx <tab>':
  (a) grouped-token binary (first commit) -> candidates added: 0   -> zsh lists the whole dir (the bug)
  (b) plain-token binary (this fix)       -> candidates added: 44  -> names render, no fallback
```

### New approach (commit `b80844d7`)
Keep the wire protocol fixed to Click's stable `plain`/`file`/`dir` triples; do **all** grouping inside the zsh template. The template splits on the file/dir directive:
- **File-union value positions** (e.g. `rbx run` solutions, which emit a `file` directive) → `compadd -d` (insertion order preserved: `@main` first, `@boca/` last).
- **Name positions** (options/subcommands, no file directive) → `_describe`, which collapses same-help aliases onto one sorted line.

Because the binary no longer emits any custom type, a **stale installed completion degrades gracefully** (names render ungrouped) instead of exploding — and reinstalling the new template restores grouping. `resolve()` is unchanged, so bash/fish output and the Typer differential parity are untouched. New guard `test_zsh_emits_only_stable_directive_types` prevents reintroducing a custom type. 844 completion tests pass.

> Verified end-to-end in real zsh under oh-my-zsh: names group + sort, solutions stay in order, no directory-listing blowup.

PR #598, opened 2026-06-12
feat(packaging): add DOMjudge problem packager

rsalesc:
## Summary

Adds `rbx package domjudge`, building a DOMjudge-importable problem zip. The layout follows the ICPC problem package format plus DOMjudge extensions, mirroring what [pol2dom](https://github.com/dario2994/pol2dom) produces (known to import cleanly):

```
domjudge-problem.ini      # short-name, name, exact fractional timelimit, contest color
problem.yaml              # limits {memory, output}, validation[, validator_flags]
problem.pdf               # the already-built main statement PDF
data/sample/, data/secret # 001.in/001.ans per-directory counters
output_validators/        # custom checker (flattened) + patched testlib.h + rbx.h
submissions/              # jury solutions by expected outcome
```

Design doc: `docs/plans/2026-06-12-domjudge-packager-design.md`.

## Key decisions

- **Checker mapping**: checkers resolving to the bundled builtins map to DOMjudge's default output validator (`wcmp`/`ncmp`/`yesno` → no flags, `dcmp` → `float_tolerance 1e-6`). Anything else — including a same-named local file, which may have been user-edited — ships as a custom validator, flattened via `flattening.build_flat_namespace` (cross-dir includes rewritten).
- **testlib patch** (`domjudge/testlib_patch.py`): DOMjudge validators speak the Kattis protocol (exit 42/43, team output on stdin, feedback dir), which vanilla testlib doesn't. Ports pol2dom's patch (from cn-xcpc-tools/testlib-for-domjudge), applied to the bundled testlib at package time; raises if an anchor is missing so a testlib upgrade fails tests, not packages. Verified locally: patched `wcmp` compiles and exits 42/43 with `judgemessage.txt` under the DOMjudge calling convention.
- **Solutions**: ACCEPTED→accepted, WA→wrong_answer, TLE→time_limit_exceeded, RTE/MLE→run_time_error (DOMjudge reports MLE as RTE); ambiguous outcomes skipped with a console note.
- **Limits**: uses the `domjudge` limits profile when saved, else package limits (not required, unlike BOCA — DOMjudge has a single per-problem TL).
- **Scope**: BATCH only; rbx COMMUNICATION pairs interactor+checker, which doesn't map onto DOMjudge's single output validator (follow-up). Build only — no upload.

## Testing

- 12 new tests in `tests/rbx/box/packaging/test_domjudge.py` (testlib patch against bundled testlib, ini/yaml content, builtin-vs-custom checker resolution, flattening, submissions mapping, sample/secret routing, full `package()` smoke).
- `tests/rbx/box/packaging` + `tests/rbx/box/completion`: 1139 passed (3 docker-compose BOCA e2e errors are pre-existing local-env failures).
- Completion spec regenerated (`mise run gen-completion-spec`); drift test green.
- Docs: new `docs/setters/packaging/domjudge.md` + index table row + mkdocs nav.


rbx/box/packaging/domjudge/packager.py
```diff
@@ -0,0 +1,241 @@
...
+
+# Builtin checkers that match the semantics of DOMjudge's default output
+# validator (possibly modulo flags). Anything else ships as a custom validator.
+_DEFAULT_VALIDATION_CHECKERS = {
+    'wcmp.cpp': None,
+    'ncmp.cpp': None,
+    'yesno.cpp': None,
+    'dcmp.cpp': 'float_tolerance 1e-6',
+}
+
```
rsalesc, line 60: please let's not resort to domjudge default output validators. let's always honor the rbx checker.

rbx/box/packaging/domjudge/packager.py
```diff
@@ -0,0 +1,241 @@
...
+# DOMjudge reports MLE as RTE by default, hence the MLE mapping.
+_SUBMISSION_DIRS = {
+    ExpectedOutcome.ACCEPTED: 'accepted',
+    ExpectedOutcome.WRONG_ANSWER: 'wrong_answer',
+    ExpectedOutcome.TIME_LIMIT_EXCEEDED: 'time_limit_exceeded',
+    ExpectedOutcome.RUNTIME_ERROR: 'run_time_error',
+    ExpectedOutcome.MEMORY_LIMIT_EXCEEDED: 'run_time_error',
+}
+
+_CPP_SUFFIXES = {'.cpp', '.cc', '.cxx', '.c++'}
```
rsalesc: please use the language kinds for the checker to identify this

rbx/box/packaging/domjudge/packager.py
```diff
@@ -0,0 +1,241 @@
...
+# validator (possibly modulo flags). Anything else ships as a custom validator.
+_DEFAULT_VALIDATION_CHECKERS = {
+    'wcmp.cpp': None,
+    'ncmp.cpp': None,
+    'yesno.cpp': None,
+    'dcmp.cpp': 'float_tolerance 1e-6',
+}
+
+# DOMjudge reports MLE as RTE by default, hence the MLE mapping.
+_SUBMISSION_DIRS = {
```
rsalesc: could you explain this better? i'm not sure i like the idea of solutions with certain outcomes just vanishing from the upload. please research how these submissions are used, when they matter, which ones we can map to their outcomes, etc. please specifically validate that domjudge (not only problem package format) supports what you're stating.

rbx/box/packaging/domjudge/packager.py
rsalesc: Done in 087f800b. Dropped the default-validator mapping entirely — the rbx checker is now **always** shipped as a custom output validator (`validation: custom`), so DOMjudge judges with exactly the checker rbx uses locally. This applies to the builtins (`wcmp`, `ncmp`, …) too. Verified locally that a builtin checker compiles and exits 42/43 under the DOMjudge calling convention after the testlib patch.

rbx/box/packaging/domjudge/packager.py
rsalesc: Done in 087f800b. Now uses `LanguageKind.CPP in environment.language_kinds(code.find_language(checker))` instead of the hardcoded suffix set, so detection derives from the actual toolchain and is robust to custom language names. Added `test_output_validators_reject_non_cpp_checker` to cover the rejection path.

rbx/box/packaging/domjudge/packager.py
rsalesc: You're right — dropping solutions was wrong. I researched how DOMjudge actually handles jury submissions (verified against the DOMjudge 9.0 source, not just the problem-package spec) and reworked it in 087f800b so **no solution is ever dropped**.

**How DOMjudge uses these:** on import it auto-judges every solution in `submissions/` and records an *expected* verdict per submission, then surfaces any mismatch on the jury **Judging verifier** page (it never blocks the import). The expected verdict comes from one of two places:
- **Directory name**, when it normalizes to a real verdict via DOMjudge's `PROBLEM_RESULT_REMAP` table (`SubmissionService.php`). That's **7** dirs, not 4: `accepted`, `wrong_answer`, `time_limit_exceeded`, `run_time_error`, `output_limit`, `compiler_error`, `no_output`.
- **An `@EXPECTED_RESULTS@:` source annotation**, which supports **multiple** acceptable verdicts (e.g. `CORRECT, TIMELIMIT`). Crucially, in a verdict-named dir DOMjudge *collapses* the annotation to the dir's single verdict; in a non-verdict dir (e.g. `mixed/`) it keeps the full list. So multi-verdict expectations must live in a non-verdict dir.

**New mapping (every outcome shipped):**

| rbx outcome | placement |
|---|---|
| ACCEPTED / WRONG_ANSWER / TIME_LIMIT_EXCEEDED / RUNTIME_ERROR | matching standard dir, no annotation |
| OUTPUT_LIMIT_EXCEEDED | `output_limit/` (DOMjudge extension dir) |
| ACCEPTED_OR_TLE | `mixed/` + `@EXPECTED_RESULTS@: CORRECT, TIMELIMIT` |
| TLE_OR_RTE | `mixed/` + `TIMELIMIT, RUN-ERROR` |
| INCORRECT | `mixed/` + all non-`CORRECT` verdicts |
| ANY | `mixed/` + all verdicts |
| MEMORY_LIMIT_EXCEEDED | `mixed/` + `RUN-ERROR, TIMELIMIT` |

**The one thing DOMjudge genuinely can't represent:** there is no memory-limit verdict and no `@EXPECTED_RESULTS@` token for it — an over-memory run surfaces as RTE (sometimes TLE). So MLE is the single lossy mapping; I encode it as `RUN-ERROR, TIMELIMIT` and documented the caveat. Everything else is exact.

Annotation comment prefix is `#` for Python, `//` otherwise. Docs/CLAUDE.md updated with the table and a note that `mixed/` solutions trigger a harmless "result does not match directory" message on import. Covered by `test_submissions_single_verdict_use_standard_dirs` and `test_submissions_ambiguous_outcomes_use_mixed_dir_with_annotation`.

rsalesc:
Rebased onto current `main` (the branch was 178 commits behind and conflicting).

**Conflicts resolved:** mkdocs nav and the packaging `CLAUDE.md` command table, both of which had gained a MOJ row on `main`; the packaging docs index table was rebuilt so the newer MOJ row fits the wider column padding this PR introduced.

**Drift checks** — a clean textual merge does not mean the 6-month-old code still fits today's APIs, so these were verified directly:

- Every API the packager calls still exists with a compatible signature (`naming.*`, `limits_info.get_limits`, `flattening.build_flat_namespace`, `code.find_language`, ...).
- The testlib patch still finds all of its anchors in the currently bundled testlib (it raises if one goes missing).
- `tests/rbx/box/packaging` + `tests/rbx/box/completion` + `lazy_cli_test`: **1581 passed, 6 skipped**, including the completion-spec drift test.

**End-to-end verification:** built a real package from a fresh `--preset default` problem with generated secret tests plus WA/INCORRECT solutions. Layout is correct (`domjudge-problem.ini`, `problem.yaml`, `problem.pdf`, `data/sample` + `data/secret`, `output_validators/`, `submissions/` routed by outcome with `mixed/` carrying the `@EXPECTED_RESULTS@` annotation). The shipped validator compiles and obeys the Kattis protocol: exit 42 on accept, 43 on reject, `judgemessage.txt` written to the feedback dir.

**One addition:** `rbx package domjudge` was missing from the checked-in CLI reference. Added by hand rather than regenerating, since the generator rewrites that whole page and it has drifted from what is committed.

Scope is unchanged: BATCH-only, build-only (no upload, no interactive problems).

rsalesc:
CI note: the failing test job is not from this PR. The 3 failures are all in `tests/rbx/box/test_timing_run_all_cli.py` and reproduce identically on a clean `main` checkout with none of these changes; bisected to efaf702f (#832). Tracked in #844.

Lint & Format passes, and the packaging/completion/lazy-CLI suites are green (1581 passed).

docs/setters/packaging/domjudge.md
```diff
@@ -0,0 +1,115 @@
...
+
+(*) DOMjudge has no memory-limit verdict — an over-memory run surfaces as a runtime
+error (sometimes a time limit). This is the one outcome that can't be expressed
+exactly.
+
+!!! note
+    Solutions in `submissions/mixed/` trigger a harmless "result does not match
+    directory" message on import; this is expected and does not block anything.
+
+### Solutions that span more than one file
```
rsalesc: Remove this section.

rsalesc: Done in 7eb37874 — section removed.

Agreed it doesn't earn its place: a solution split across headers of its own just works, so there is nothing for a setter to do differently, and the two cases that *don't* work report themselves at packaging time with the offending file named. The mechanism stays documented for contributors in the packaging `CLAUDE.md`.

Also in this round, from your earlier note about `extra_roots` (f69844b8): dropping it turned out to matter more than it looked. `header.get_header()` returns the *relative* path `rbx.h`, so its `.parent` is `.` — passing it silently made the **package root** an include root for every solution. That is what my `lib.h` test had been passing on: it put the header at the package root with the solution in `sols/`, which the compiler would not resolve either. Both header tests now use a real sibling, and a solution reaching for `rbx.h`/`testlib.h` is refused rather than inlined (`rbx.h` reads a vars file the judge lacks, so inlining it would only trade a compile error for a runtime one).

Separately, verifying this against a live DOMjudge 9.0.0 turned up that the amalgamator stamps each inlined file with the setter's **absolute path** — it ships in MOJ packages too. Filed as #855; not fixed here.

PR #614, opened 2026-08-08
feat(presets): problem template variants (#264)

### LINKED ISSUE #264: Add variants for problem templates in presets
Useful for interactive problems, for instance

rsalesc:
Closes #264 — problem template variants in presets.

## Why

A preset declared exactly one problem template, so every problem started as a batch problem. Interactive and subtask-scored problems differ in ways `rbx/box/schema.py` treats as **mutually exclusive**:

- **Interactive**: `type: communication` + an `interactor:` and **no** `checker` (`check_checker_and_interactor_for_task_type` rejects the other combinations), plus an `interactor.cpp`, an interaction-loop solution, and a statement with an Interaction section.
- **Points**: `scoring: points` with `score:`/`dependencies:` on groups and expected score ranges on solutions (`check_scoring_fields` rejects all three when scoring is binary).

So starting an interactive problem meant copying the batch template and rewriting half of it — exactly the work a preset exists to remove.

## What this does

```yaml
# preset.rbx.yml
problem: "problem"
problemVariants:
  - id: interactive
    path: "problem-interactive"
    description: "Communication task with a testlib interactor"
```

```bash
rbx create guess -v interactive     # or omit -v and pick from a list
```

- A variant is a **complete sibling template directory**, so it stays a runnable package you can `rbx build` inside — the workflow the preset docs already recommend. That testability is what ruled out base+overlay dirs and a single Jinja-conditional template.
- Per-variant `tracking` / `libraries` / `expansion` merge over the preset's shared lists (variant wins, keyed by path / library name / needle). `default` is a reserved id meaning the canonical template. Variants-only presets (no canonical) are legal. `contestVariants` works identically.
- `--variant/-v` on `rbx create`, `rbx contest add`, `rbx contest create`. Omitted + variants declared + TTY → picker; off-TTY → canonical. **A preset with no variants behaves exactly as before** — no prompt, no new output.
- The choice is recorded as `variant:` in `.preset-lock.yml` and is **fixed at creation**. `rbx presets sync` resolves against that template and hard-errors if it disappeared rather than falling back to canonical — falling back would overwrite an interactive problem's tracked assets with batch ones. `variant: null` means canonical, so every existing lock keeps working untouched.
- `rbx presets ls` lists each kind's templates with id, description, path and status (`in use` / `missing`).
- The bundled `default` preset ships an **`interactive` variant** — the guessing game from the interactors docs. Verified end to end: `rbx create -v interactive` → `rbx build` → `rbx run` judges the reference solution accepted on every test.

## Bugs found and fixed along the way

None of these were in the plan; each was found by review or by testing:

- **`rbx presets lint`/`fix` could reformat a package outside the preset.** `fix_package` recursed into contest template dirs, and `find_package` walks *up* the tree without stopping at `preset.rbx.yml` — so a template dir with no `contest.rbx.yml` escaped the preset and reformatted whatever unrelated package sat above it. Pre-existing; variants would have multiplied it by the number of contest variants. Regression test included (it fails loudly with the guard removed).
- **Declared paths had no containment check.** `PackageVariant.path`, `Preset.problem`/`contest`/`env`, `Library.dest` and `TrackedAsset.path` accepted absolute paths (silently discarding the preset root) and `../..` escapes — while `copy_preset_file` already refused escaping *symlinks*. `Library.dest` writes into the user's package. Now validated lexically at the schema, with a real-path containment check in the resolver for symlinked template dirs.
- **Lock/install drift, in four places.** `_sync`, the create path, `rbx contest create`/`init`, and the hidden `rbx presets lock` each resolved the template independently of the install — harmless while everything was canonical, but each would have silently recorded the wrong variant. Installs now return the template they resolved and every lock-writing site threads it.
- **Template linting skipped variants entirely**, so an interactive template's `interactor.cpp` would never be linted.
- **Stale test fixtures were silently disabling 21 tests.** Preset fixtures omitted `min_version` and inherited a `0.14.0` default, which reads as a breaking change against rbx 1.0.0 — so the compatibility check exited before any assertion ran. This branch originally pinned the fixtures as a workaround; #613 has since fixed the root cause by bumping the schema default, so **those pins have been reverted** and this branch now defers to it entirely.

## Testing

370 preset tests, plus creation/contest/linting/e2e. New e2e fixture `default-preset-interactive` exercises the interactor end to end (`build` → `run` → `st b`). Full suite is at its pre-existing baseline — remaining failures are the known local C++/sandbox families, none in touched code.

Several guards were verified by falsification rather than by assertion alone: the sync tests were re-run with `_sync` forced to resolve canonical (both fail, with the data-loss visible in captured output), the merge helper was mutated four ways, and the linting guard was removed to confirm the ancestor package really does get reformatted without it.

## Synced with main

Merged `main` through cadebeda. Three things the merge required, all resolved:

- **#613 (min_version)** — adopted at the root; this branch's fixture-level pins are reverted, since their comments claimed a staleness that no longer exists.
- **#617 (versioned schemas)** — kept `_pin_schema_header` alongside this branch's narrowed `materialize_libraries` signature (the one conflict). Verified: a package created with `-v interactive` gets the versioned `rbx-schemas/1.0/Package.json` header, same as the canonical template.
- **#616 (getVar path segments)** — the canonical `rbx.h` changed, so the interactive variant's copy was re-synced. **The drift test caught this**, which is the reason it exists.

Design and plan: `docs/plans/2026-08-08-preset-problem-variants-design.md`, `docs/plans/2026-08-08-preset-problem-variants.md`.


docs/setters/presets/index.md
```diff
@@ -90,6 +90,171 @@ and the `name` field of the `problem.rbx.yml` file, which will be changed to mat
...
+rbx run
+```
+
+Once it works, declare it in `problemVariants:` as shown above.
+
+!!! tip
+    The bundled `default` preset already ships an `interactive` variant (the guessing game
+    from [Interactors](../grading/interactors.md)). If you created your preset from it with
+    `rbx presets create`, you already have a working example to crib from in
+    `problem-interactive/`.
```
rsalesc: Please remove this note, doesn't make sense to have this here.

docs/setters/presets/index.md
```diff
@@ -90,6 +90,171 @@ and the `name` field of the `problem.rbx.yml` file, which will be changed to mat
...
+
+#### Selecting a variant
```
rsalesc, line 149: The interactive way should be the default way we explain to users, and the `-v` flag is the power user way.

docs/setters/presets/index.md
```diff
@@ -90,6 +90,171 @@ and the `name` field of the `problem.rbx.yml` file, which will be changed to mat
...
+
+- **Plain copies**: just keep a copy of the file in each template directory. Simple, and
+  what the bundled `default` preset does.
+- **Symlinks inside the preset**: point a file in one template at a file elsewhere in the
+  preset. When the package is created, {{rbx}} rewrites such a symlink into a relative link
+  into the installed package's `.local.rbx/`, so the created package stays self-contained
+  and the file keeps following the preset.
+
+#### The variant is fixed at creation
+
```
rsalesc: Please remove this section, it doesn't make sense.

docs/setters/presets/index.md
rsalesc: Removed in 6b947c76.

docs/setters/presets/index.md
rsalesc: Flipped in 6b947c76. "Selecting a variant" now opens with a bare `rbx create --preset your-preset` and the picker it shows (transcript included, matching the real prompt strings), notes that a preset without variants never prompts, and moves `-v/--variant` into a "Skipping the prompt" tip alongside the non-TTY fallback.

Also realigned two spots that still led with the flag: the YAML comment in the intro ("selectable with `-v/--variant`" → "offered when a problem is created") and the contest example.

docs/setters/presets/index.md
rsalesc: Removed in 6b947c76, along with the trailing reference to it in the contest section ("the lock entry, the sync behavior — applies unchanged").

Flagging one thing that section was the only documentation for: if a preset drops a variant that an existing package is locked to, `rbx presets sync` fails instead of falling back to the canonical template. It's covered by tests, and the error message itself explains the situation and points at `.preset-lock.yml`, so nothing is silently lost — just say the word if you'd like it written down somewhere shorter, otherwise I'll leave it to the error message.

PR #617, opened 2026-08-08
feat(schemas): versioned JSON schemas with min_version-pinned URLs

rsalesc:
## Summary

Schemas are currently published from a single unversioned URL that always tracks `main`, so a package pinned to an older rbx is validated against a schema newer than the tool reading it. This ships versioned schemas and makes rbx write a version-pinned `# yaml-language-server` header.

- **Hosting** — a new `rsalesc/rbx-schemas` Pages repo, written additively by the release (locally *and* from CI) (`<major>.<minor>/<Model>.json`, plus `latest/` and `index.json`). Additive writes never collide with the docs `gh-pages` branch, which `mkdocs gh-deploy --force` fully replaces.
- **Pinning** — the URL carries the `min_version` of the active preset (the compatibility floor the package promises), resolved by the new `rbx/box/schema_urls.py`. Files with no preset in scope pin to the installed version.
- **Forward tolerance** — published schemas drop `additionalProperties: false`, so a field added after the pinned minor is not flagged. rbx itself still hard-errors on unknown keys at load time.
- **Write paths** — `model_to_yaml`, `rbx fix` (this re-enables `fix_language_server`, disabled since it was added), and preset materialization, so third-party presets that hardcode a URL get normalized too.
- **Backwards compatible** — the existing unversioned URLs keep being published on every docs deploy. No file in the wild changes meaning.

Design and plan: `docs/plans/2026-08-08-versioned-schemas-design.md`, `docs/plans/2026-08-08-versioned-schemas-plan.md`.

## Publishing from both release entrypoints

Releases here happen manually from the CLI (`cz bump` locally, sometimes before anything reaches GitHub), so schema publishing cannot live only in the tag workflow. `scripts/publish_schemas.py` is the single code path:

| Entrypoint | Command |
| :--- | :--- |
| Manual release | `mise run release` (bump + PyPI + schemas), or `mise run publish-schemas` alone |
| Tag pushed to GitHub | the `publish-schemas` job, which calls the same script with `--dir` |

- The version defaults to the installed one — exactly what `cz bump` just wrote to `rbx/__version__.py` — so publishing right after a bump targets the version being released.
- Publishing is **idempotent**: whichever entrypoint runs second sees identical content and commits nothing. Both may fire safely.
- Prereleases are skipped unless `--allow-prerelease`.
- The script refuses to run against anything that is not a schemas checkout. It writes files, runs `git add -A` and pushes, so an empty `--dir` (which resolves to the current directory) or a stray path would otherwise sweep unrelated work into a commit.

## Notable decisions

- **The pin floor is `1.0`**, the minor the bundled preset already declares. Note `pyproject.toml` lists that preset's `min_version` in commitizen's `version_files`, so each bump rewrites it and packages created after a release pin to that release's minor automatically. The floor can never exceed the installed version: a preset above the installed version is rejected outright (`OUTDATED`) and one below the floor falls back, so a higher floor would be unreachable. This also means **no compatibility-breaking `min_version` bump** and it can ship in a patch. A guard test asserts that once the installed minor reaches the floor, the bundled preset declares a `min_version` at or above it — so a future version bump cannot leave pinning silently inert.
- **Minor granularity means mutable URLs**: a patch re-publishes its own minor, so `1.4` tracks the newest `1.4.x`. That is the correction path for a bad publish, and the reason a pin is a compatibility contract rather than a reproducibility guarantee.
- **Forward tolerance does not cover enum values.** A value added after the pinned minor is still rejected by the older schema; there is no way to express "and anything added later" for an enumeration. The fix is raising the preset's `min_version`, which is the honest signal. Documented.
- `fix_language_server` had a latent bug: it re-inserted the header only after a `---` line, so a file without a document marker lost its header entirely. Fixed, with a test.
- **The bundled preset resources keep the unversioned URL.** Commitizen rewrites the preset's `min_version` on every bump but cannot rewrite a `major.minor` URL path, so a committed pin would go stale at the next release. Headers are normalized when a package is materialized (problem, contest, and now preset packages) or by `rbx fix`.
- `schema_urls.preset_min_version` is registered in `rbx.testing_utils.clear_all_functools_cache`, per the cache rule in `rbx/box/CLAUDE.md`.

## Status: live

`rsalesc/rbx-schemas` exists with Pages serving `main` at `/`, and schemas for **1.0** are published and verified:

| URL | Result |
| :--- | :--- |
| `index.json` | 200, `application/json` — `{"latest":"1.0","versions":["1.0"]}`, 8 models |
| `1.0/Package.json` | 200, `application/json`, 65,873 bytes, byte-identical to the local generator, no `additionalProperties: false` |
| `latest/Package.json` | 200, identical to `1.0/` |

A package materialized from the bundled preset came out with `$schema=https://rsalesc.github.io/rbx-schemas/1.0/Package.json`, which fetches 200 — so pinning and hosting agree. Re-running the publisher correctly reported no changes.

**Optional, not required:** adding a `SCHEMAS_DEPLOY_TOKEN` secret (fine-grained PAT or deploy key with write access to the schemas repo) turns on the CI backstop. Without it that job skips quietly and never fails a release; the manual `mise run release` path pushes over the normal git credential helper.

## Test Plan

- [x] `tests/rbx/box/test_schema_urls.py` — pin resolution: preset floor, nested problem dir, below-floor fallback, no preset, malformed YAML (silent), incompatible preset (does not raise), caching
- [x] `tests/rbx/box/linters/test_language_server_header.py` — replace/idempotency/no-marker/foreign-URL-untouched/no-duplication
- [x] `tests/rbx/box/presets/test_preset_schema_headers.py` — pinned on problem+contest creation, below-floor fallback, real installed version without mocking, floor-reachability guard
- [x] `tests/rbx/box/test_schema_export.py` — model list matches, relaxation keeps dict-typed `additionalProperties` schemas, site layout, merge with existing versions, no `latest` demotion, numeric version sort, unrelated dirs ignored, prerelease skip, version defaulting, publish idempotency, and the three refusal guards (empty dir / source checkout / non-git)
- [x] Local publish dry-run end to end against a throwaway bare remote: clone → build → commit → push, then a re-run that correctly published nothing
- [x] 462 passed across presets / contest / linters / schema suites / `test_utils`
- [x] e2e: 27 passed, 6 failed — the same 6 IDs fail on `origin/main` (verified in a clean worktree)
- [x] `mkdocs build` succeeds; still emits `site/schemas/*.json`
- [x] Manual: materialized the bundled preset, confirmed `$schema=.../rbx-schemas/1.0/Package.json`; reverted the header to the unversioned URL and ran `rbx fix` twice — re-pinned, then no change
- [x] Live: `index.json`, `1.0/*.json` and `latest/*.json` all serve 200 as `application/json`; served `Package.json` byte-matches the local generator and carries no `additionalProperties: false`
- [x] Live: a package materialized from the bundled preset pins to `/1.0/Package.json`, which fetches 200; re-publishing is a no-op
- [ ] Yours: open a generated `problem.rbx.yml` in VS Code and confirm completion works with no "unable to load schema" error


docs/internal/schemas.md
```diff
@@ -0,0 +1,97 @@
+# JSON schemas
```
rsalesc, line 1: not sure we need this page. it's ok to keep it, but maybe in a section "Internal reference" that for now is hidden. We'll add more interesting stuff in there in the future.

rbx/box/internal/schema_export.py
```diff
@@ -0,0 +1,31 @@
+import pathlib
```
rsalesc, line 1: let's maybe move this and the other added schema utilities into an "rbx/box/internal" folder as this has more to do with the rbx development itself than with "rbx" functionalities?

docs/internal/schemas.md
rsalesc: Moved to `docs/internal/schemas.md` under a commented-out `Internal reference` nav section in `mkdocs.yml`. The page still builds and is linkable at `/internal/schemas/` — it just isn't in the nav, same as `docs/plans/*` today. Uncommenting those two lines is all it takes to show the section once there's more in it.

rbx/box/internal/schema_export.py
rsalesc: Moved `schema_export.py` and `schema_publish.py` to `rbx/box/internal/` (with `tests/rbx/box/internal/` mirroring it). Nothing in the CLI reaches either one — only the docs build, the release tasks and CI.

I kept `schema_urls.py` in `rbx/box` on purpose: it is runtime rbx functionality, not dev tooling. `rbx/utils.py:model_to_yaml` and `rbx/box/linting.py` call it to stamp the `# yaml-language-server` header into user packages, and it's registered in `clear_all_functools_cache`. Say the word if you'd rather it move anyway.

Two pre-existing neighbours are arguably internal too — `rbx/box/dump_schemas.py` and `rbx/box/dump_cli_docs.py`, both mkdocs `gen-files` scripts. I left them alone to keep this diff to what the PR added; happy to move both as a follow-up.

PR #619, opened 2026-08-09
feat: per-group vars for validators

rsalesc:
## Why

A validator branching on the test group is dead code:

```cpp
std::string group = opt<std::string>("group", "");
if (group == "sub2") { ensuref(...); }   // never runs
```

rbx *does* pass `--group`, but testlib's `opt<>()` reads a map filled only by
`prepareOpts()`, and `registerValidation` never calls it (only `registerGen` does,
`testlib.h:4627`). The check evaporates and the build goes green.

Adding `prepareOpts` swaps one trap for another: `opt(key, default)` arms testlib's
unused-opt check, and rbx injects every package var as `--{k}={v}`, so every unread
var becomes `FAIL Opts: unused key 'N.max'`.

The deeper problem is that group-name branching fails **silently** whenever a group
is renamed. So this PR makes the constraint a *parameter* instead of a branch.

## What

```yaml
vars:
  AB: {min: -200, max: 200}

testcases:
  - name: "sub2"
    vars: {AB: {min: 0}}     # deep-merged; AB.max stays 200
  - name: "sub4"             # no override: package defaults
```

```cpp
// validator.cpp — no group branching at all
int b = inf.readInt(getVar<int>("AB.min"), getVar<int>("AB.max"), "B");
```

`getVar`'s signature does not change: an existing validator becomes group-aware the
moment a `vars:` block is added to a group.

Statements read the same values, resolved: `problem.groups.sub2.vars.AB.min`. A group
overriding nothing still renders the inherited value, so a subtasks table cannot
silently render blanks.

**Validators only.** Checkers and interactors follow the Kattis convention of taking
context from the input/answer files (both judge-controlled data in rbx); testlib's
`registerInteraction` cannot accept extra flags at all — it records `--group` but never
strips it, and `argv[3..5]` are hard-wired to answer/report/`-appes`.

## How the group reaches `rbx.h`

`rbx.h` parses `--group` itself, with no dependency on testlib and no call required in
`main()`, by reading the process's own command line: `_NSGetArgv` (Apple),
`/proc/self/cmdline` (Linux), `__argc`/`__argv` (Windows), empty elsewhere.

A `__attribute__((constructor))` hook taking `(argc, argv, envp)` was tried and
**rejected**: at `-O2` GCC folds it into the TU's argument-less static-init thunk when
the TU has a dynamic initializer — and every testlib TU has several — silently losing
argv. It works on Apple clang and at `-O0`, which is what makes it a trap. The design
doc records the evidence.

## Notable fixes found along the way

- `getVar<bool>` returned `true` for a **missing** variable. The fallback was
  `opt = getIntVar(name) != 0`, and comparing a disengaged `std::optional` with `0`
  yields `true`, so the `throw` below was unreachable. Pre-existing.
- `rbx.h` sentinel substitution used `str.replace`, a silent no-op on a missing or
  renamed sentinel — which would ship a header that compiles and resolves nothing.
  Now asserts exactly-one-match.
- Reserved testlib flag names (`group`, `testset`, `testCase`, …) are now rejected as
  top-level var names: a var named `group` emitted `--group=<value>` *before* the real
  flag.
- `prepareOpts` removed from the default preset validator (and its three copies in
  docs/fixtures), where it enabled nothing but armed the unused-opt trap.

## Verification

707 tests pass across schema/fields/header/package/statements/contest/presets/linters.
Header tests do real `g++` compile-and-run, including per-group arms, type-changing
overrides, group-only vars, sorted-arm order, guard emission and `-Wextra` cleanliness.
Two e2e fixtures cover the pipeline end to end, including behavioural coverage that the
hit-bounds report stops merging across groups.

## Known gaps, recorded rather than glossed

- The Windows `__argc`/`__argv` branch has **never run on Windows**; `/proc/self/cmdline`
  has never run on real Linux locally. Linux CI is the gate on the latter.
- Subgroups deliberately cannot declare `vars` (`--group` carries only the top-level name).
- Bool vars go out on argv as Python `True`/`False`, so `opt<bool>` cannot read them —
  documented; changing the rendering is CLI-visible and left as a maintainer call.
- Per-group vars do not survive Polygon export (Polygon re-validates with a single
  ungrouped testset). A narrowing override is harmless; a widening one is not.

## Separate pre-existing finding, not fixed here

`rbx build` **exits 0 even when validation fails** — `builder.build()` returns `False`
(`builder.py:72`) but `cli.py:314` discards it. A CI pipeline running `rbx build` goes
green despite a failing validator. Not touched here because changing CLI exit semantics
could break existing scripts.


docs/setters/reference/package/index.md
```diff
@@ -394,6 +394,76 @@ You can pass variables to validators in two different ways.
...
+    !!! warning "Booleans are not readable from the command line"
+
+        Boolean variables are rendered on the command line with Python's spelling --
+        `--flag=True` / `--flag=False` -- while the generated `rbx.h` emits C++'s
+        `true` / `false`. {{testlib}}'s `opt<bool>` accepts only `true`, `false`, `0`
+        or `1` and fails the run on anything else, so read boolean variables with
+        `getVar<bool>` (or compare the raw string yourself).
+
+{{rbx}} additionally passes `--group <name>` naming the top-level testgroup being validated
+(and `--testOverviewLogFileName`, which {{testlib}} uses to report the bounds your tests hit).
```
rsalesc: No need, please omit, implementation detail.

docs/setters/reference/package/index.md
```diff
@@ -394,6 +394,76 @@ You can pass variables to validators in two different ways.
...
+
+!!! danger "`opt()` does not work in a {{testlib}} validator"
```
rsalesc: Please omit, no need to keep this.

docs/setters/reference/package/index.md
```diff
@@ -394,6 +394,76 @@ You can pass variables to validators in two different ways.
     # Use variables...
     ```
 
+    The values passed here are already resolved for the testgroup being validated, so an
+    `argv`-parsing validator and a `getVar` one see the same numbers and strings.
+
+    !!! warning "Booleans are not readable from the command line"
```
rsalesc: Please remove this bit, I'll try to fix the issue.

docs/setters/verification/validators.md
```diff
@@ -280,9 +280,167 @@ testcases:
       - path: 'straight-validator.cpp'
 ```
 
+## Varying constraints per test group
 
+Subtasks usually differ only in *how large* the input may get, not in what a valid
```
rsalesc: This is not necessarily true for competitive programming problems (think connected graphs). But still, it's possible to solve this with a single validator.

Please simply don't invent a reason. Let's just say this is recommended.

docs/setters/verification/validators.md
```diff
@@ -280,9 +280,167 @@ testcases:
...
+validated to every validator run, and the generated `rbx.h` resolves `getVar` against
+it. A test that is legal in `large` and too big for `small` now fails in `small`, and
+`rbx build` reports it.
+
+A few details worth knowing:
+
+- **The merge is deep, at the leaf.** `vars: {N: {max: 50}}` overrides `N.max` and
+  leaves every sibling (`N.min`, and every other variable) alone. You never have to
+  restate the constraints you are not changing.
+- **The keys do not have to exist at package level.** A variable that is only
```
rsalesc: Please remove this.

docs/setters/verification/validators.md
```diff
@@ -280,9 +280,167 @@ testcases:
...
+  meaningful inside one group can be declared in that group alone. Reading it from
+  another group fails loudly with the usual `getVar` error (*Variable `X` ... could
+  not be found*), rather than quietly returning something.
+- **Only top-level testgroups can declare `vars`.** Subgroups cannot: {{rbx}} only
+  tells the validator the top-level group name, so a subgroup override would have no
+  way to be selected at runtime.
```
rsalesc: Please remove the explanation of why this is true, simply keep just the fact.

docs/setters/verification/validators.md
```diff
@@ -280,9 +280,167 @@ testcases:
...
+- The group's resolved values are also what {{rbx}} passes on the validator command
+  line as `--{name}={value}`, so a validator written in another language that parses
+  `argv` sees the same numbers and strings as a C++ one using `getVar`. Booleans are
```
rsalesc: Remove the booleans exception, it's not.

docs/setters/verification/validators.md
```diff
@@ -280,9 +280,167 @@ testcases:
...
+  the exception: they are rendered with Python's spelling (`--flag=True`), not C++'s,
+  so `opt<bool>` cannot read them -- use `getVar<bool>`.
+
+### Reading the group name
```
rsalesc: This is too brittle, please remove this section.

docs/setters/verification/validators.md
```diff
@@ -280,9 +280,167 @@ testcases:
...
+
+    A `if (group == "small")` branch is only as good as the group's name. Rename the
+    group, split it in two, or reuse the validator in another problem, and the branch
+    silently stops matching — no error, the check simply evaporates and the build
+    stays green.
+
+    A `vars` override travels *inside* the group entry, so renaming the group cannot
+    detach it.
+
+### `opt()` does not work in a validator
```
rsalesc: Please remove this section, do not explain testlib.

docs/setters/verification/validators.md
```diff
@@ -280,9 +280,167 @@ testcases:
...
+
+There is no equivalent for checkers and interactors, deliberately. They follow the
+convention {{rbx}} shares with [Kattis](https://github.com/Kattis/problem-package-format/issues/393):
+whatever context they need belongs in the input or answer file, both of which are
+judge-controlled data you generate yourself. If a checker has to know it is judging a
+subtask, encode that in the answer file.
+
+For interactors, it is also a hard {{testlib}} constraint rather than a preference:
+`registerInteraction` hard-wires `argv[3..5]` to the answer file, the report file and
+`-appes`, so appending any extra flag corrupts its argument parsing.
```
rsalesc: Remove section.

docs/setters/verification/validators.md
```diff
@@ -280,9 +280,167 @@ testcases:
...
+
+### Showing the constraints in the statement
```
rsalesc: This section seems out place. If it should be put somewhere, it definitely should be put close to wherever we already explain how vars can be shown in statements, with the addendum this is points specific / for problems with custom groups.

docs/setters/cheatsheet.md
```diff
@@ -234,6 +234,38 @@ testcases:
...
+    # Define tests... # (2)!
+```
+
+1. Overrides `N.max` for this group only. The merge is leaf-by-leaf, so `N.min` stays
+   at the package-level `1`. The validator is untouched: `getVar<int>("N.max")` returns
+   `50` here and `1000` elsewhere.
+
+2. No override, so the package-level values apply.
+
+!!! danger "`opt()` does not work in a {{testlib}} validator"
```
rsalesc: Remove this note.

PR #635, opened 2026-08-12
fix(header): serve getVar for every built-in integer spelling

### LINKED ISSUE #634: Mismatching rbx.h `getVar<>` and testlib's `readLong` over long longs
Basically, getVar is specialized for `int32_t` and `int64_t`. In 64-bit systems, usually, `int64_t == long`, which means `getVar<long long>` is not available, and `int64_t` must be used.

Unfortunately, testlib `readLong` uses `long long` specialization, and thus at the end of the day, both interfaces mismatch. Since we don't own testlib and it's a widespread library, we should add support for `getVar<long long>` for proper compatibility.

rsalesc:
Closes #634.

## The bug

`rbx.h` defined `getVar<T>` through one explicit function specialization per fixed-width alias (`int32_t`, `uint32_t`, `int64_t`, `uint64_t`). Which *built-in* spelling those aliases name is platform-dependent:

| platform | `int64_t` is | so `getVar<...>` is missing for |
|---|---|---|
| LP64 Linux | `long` | `long long` |
| macOS / Windows | `long long` | `long` |

Setters hit this through testlib, whose `readLong()` returns a plain `long long`, so `inf.readLong(getVar<long long>("N.min"), ...)` failed to link on Linux.

## The fix

Dispatch through a trait-matched `VarReader<T, Enable>` class template instead of alias-keyed function specializations:

- partial specialization for `std::is_integral_v<T> && !std::is_same_v<T, bool>`
- partial specialization for `std::is_floating_point_v<T>`
- full specializations for `bool` and `std::string`

**On the conflict risk:** this is why the fix isn't just "add a `long long` specialization". That specialization would *redefine* `int64_t`'s on macOS and Windows, breaking the platforms that work today, and no preprocessor test can reliably tell which spelling an alias resolves to. With trait matching, one definition serves every spelling, so the conflict is structurally impossible rather than guarded against by hand. As a bonus, `VarReader`'s member functions are implicitly inline, unlike the explicit function specializations they replace.

## Behavior notes

- Every integer spelling now works, including `short`, `char`, `long`, `long long` and the unsigned counterparts, all range-checked against the `int64_t` the tables store. Error messages still name the fixed-width alias (`does not fit in int32_t`) so they stay stable across platforms.
- Range checking is now uniform: reading a negative value into `uint64_t` / `unsigned long long` throws instead of silently wrapping. `uint32_t` already behaved this way; `uint64_t` was the outlier.
- An unsupported `T` now fails with a `static_assert` message instead of an undefined-symbol link error.
- `if constexpr` guards the width comparisons so no always-true comparison is emitted, keeping the header clean under `-Wextra -Werror` (`-Wtype-limits`).

## Testing

- New `test_get_var_supports_every_integer_spelling`: compiles and runs a program reading vars through `long long`, `unsigned long long`, `long`, `unsigned long`, `int`, `unsigned int`, `short` and `int64_t` in one TU, plus overflow and negative-into-unsigned throws. It fails on `main` (undefined symbols).
- New `test_get_var_rejects_unsupported_types_at_compile_time`.
- Manually compiled a testlib validator using `inf.readLong(getVar<long long>(...), ...)` together with `int64_t`, `long` and `unsigned long long` reads in the same TU.
- `tests/rbx/box/test_header.py`: 56 passed. The 5 `TestGroupVars` failures and the `validators_test.py` failures on this machine are pre-existing — they reproduce identically with these changes stashed.

### DIFF
--- a/docs/setters/variables.md
+++ b/docs/setters/variables.md
@@ -52,8 +52,11 @@ In this section, we go through the different ways to use variables in each of th
 that were defined in the `problem.rbx.yml` file, right at the root of your package.
 
 This header exposes a function `getVar<T>(name)` that can be used to get the value
-of a variable as a `T`-typed object. There are 4 overloads for this function:
-`getVar<bool>(name)`, `getVar<int>(name)`, `getVar<float>(name)` and `getVar<std::string>(name)`.
+of a variable as a `T`-typed object. `T` can be `bool`, `std::string`, any
+floating-point type (`float`, `double`) or any integer type -- both the plain
+spellings (`int`, `long`, `long long` and their unsigned counterparts) and the
+fixed-width aliases (`int32_t`, `int64_t`, ...). Reading an integer that does not
+fit in the requested type raises an error instead of silently truncating it.
 
 Nested variables can be addressed either by their full dotted name, or by passing
 one segment per argument -- the two calls below are equivalent:
--- a/docs/setters/variables.md
+++ b/docs/setters/variables.md
@@ -53,14 +53,7 @@ that were defined in the `problem.rbx.yml` file, right at the root of your packa
 
 This header exposes a function `getVar<T>(name)` that can be used to get the value
 of a variable as a `T`-typed object. `T` can be `bool`, `std::string`, any
-floating-point type (`float`, `double`) or any integer type -- both the plain
-spellings (`int`, `long`, `long long` and their unsigned counterparts) and the
-fixed-width aliases (`int32_t`, `int64_t`, ...). Reading an integer that does not
-fit in the requested type raises an error instead of silently truncating it.
-
-Character types (`char` and friends) are deliberately not supported: it would be
-ambiguous whether `getVar<char>("x")` should give you the number or the letter.
-Read the variable into an explicit integer type instead.
+floating-point type (`float`, `double`) or any integer type.
 
 Nested variables can be addressed either by their full dotted name, or by passing
 one segment per argument -- the two calls below are equivalent:
--- a/docs/setters/variables.md
+++ b/docs/setters/variables.md
@@ -58,6 +58,10 @@ spellings (`int`, `long`, `long long` and their unsigned counterparts) and the
 fixed-width aliases (`int32_t`, `int64_t`, ...). Reading an integer that does not
 fit in the requested type raises an error instead of silently truncating it.
 
+Character types (`char` and friends) are deliberately not supported: it would be
+ambiguous whether `getVar<char>("x")` should give you the number or the letter.
+Read the variable into an explicit integer type instead.
+
 Nested variables can be addressed either by their full dotted name, or by passing
 one segment per argument -- the two calls below are equivalent:
 
--- a/rbx/resources/templates/rbx.h
+++ b/rbx/resources/templates/rbx.h
@@ -7,6 +7,7 @@
 #include <stdexcept>
 #include <string>
 #include <string_view>
+#include <type_traits>
 #include <vector>
 
 #if defined(__APPLE__)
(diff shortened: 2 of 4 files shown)


rbx/resources/templates/rbx.h
```diff
@@ -230,62 +228,100 @@ template <> uint64_t getVar<uint64_t>(std::string name) {
...
+};
+
+template <typename T>
+struct VarReader<T, std::enable_if_t<std::is_floating_point_v<T>>> {
+  static T read(const std::string &name) {
+    auto opt = getFloatVar(name);
+    if (!opt.has_value()) {
+      auto intOpt = getIntVar(name);
+      if (intOpt.has_value()) {
+        opt = (float)intOpt.value();
```
rsalesc: Should this be the actual type T instead of float?

rbx/resources/templates/rbx.h
```diff
@@ -230,62 +228,100 @@ template <> uint64_t getVar<uint64_t>(std::string name) {
...
+};
 
-template <> std::string getVar<std::string>(std::string name) {
-  auto opt = getStringVar(name);
-  if (!opt.has_value()) {
-    auto intOpt = getIntVar(name);
-    if (intOpt.has_value()) {
-      opt = std::to_string(intOpt.value());
+template <typename T>
+struct VarReader<
```
rsalesc: Can we disable reading chars? I think it's too ambiguous whether the expected input is a number or a character, so I'd rather not provide a specialization for it.

rbx/resources/templates/rbx.h
```diff
@@ -180,48 +181,45 @@ std::optional<bool> getBoolVar(std::string name) {
...
-template <> int64_t getVar<int64_t>(std::string name) {
-  auto opt = getIntVar(name);
-  if (!opt.has_value()) {
-    throw std::runtime_error("Variable " + name +
-                             " is not an integer or could not be found");
+// Every integer var is stored as int64_t, so reading one into a narrower or
+// unsigned type has to be range-checked. Written with `if constexpr` so that no
+// always-true comparison is emitted for the widest types, which -Wtype-limits
+// would reject under -Werror.
+template <typename T> bool intFits(int64_t value) {
```
rsalesc, line 222: I'm curious about the correctness of this function for unsigned 64 bits. Since the input is signed 64 bit, couldn't it be the case that the input is read and instantly overflows it?

rsalesc: Good catch — this was a regression I introduced, and it's now fixed in 314b68a0.

`header.py` accepts vars up to `2^64-1` (`check_int_bounds`) and emits them as `static_cast<int64_t>(...)`, so everything above `INT64_MAX` is stored as its two's-complement pattern. Reading such a var back through `uint64_t` used to work precisely because the old specialization did no range check at all. My uniform "negative doesn't fit an unsigned type" rule made those vars unreadable.

Unsigned reads now reinterpret the stored pattern (`static_cast<uint64_t>(value)`) and only range-check against the target's max, so:

- `uint64_t` / `unsigned long long` always succeed, recovering the original value — the old behavior;
- `uint32_t` with a stored `-1` still throws, since the reinterpretation doesn't fit — also the old behavior;
- the error message now prints the value as the requested type reads it, so an unsigned read doesn't quote a negative.

The pattern is all the table keeps, so a value above `INT64_MAX` and the negative it wraps to are genuinely indistinguishable — reading `2^64-1` as `int64_t` hands back `-1`. That's pre-existing, and the new test pins it explicitly rather than pretending otherwise.

Writing the test for this surfaced a related emitter bug worth flagging: `static_cast<int64_t>(18446744073709551615)` has no signed literal spelling and warns (`-Wimplicitly-unsigned-literal`), so with `-Werror` a var above `INT64_MAX` failed to compile the whole package. `INT64_MIN` had the same problem from the other end. Both are fixed in `_int_repr` (`...ULL` and `INT64_MIN` respectively), with a test covering both ends of the range.

rbx/resources/templates/rbx.h
rsalesc: Agreed — done in 314b68a0.

Character types (`char`, `signed char`, `unsigned char`, `wchar_t`, `char16_t`, `char32_t`, and `char8_t` where the compiler has it) are now excluded from the integral specialization and routed to a dedicated one that only carries a `static_assert`, so the failure names the reason instead of falling through to the generic "unsupported type" message:

> getVar<T>: character types are ambiguous -- it is unclear whether the var should be read as a number or as a letter. Read it into an explicit integer type, such as int or int64_t, instead

Covered by `test_get_var_rejects_char_types_at_compile_time`, and noted in the docs.

rbx/resources/templates/rbx.h
rsalesc: You're right that the `float` there was wrong — fixed in 314b68a0.

The cast couldn't literally be `T` in place, because `opt` is the `std::optional<float>` that `getFloatVar` returns, so assigning back into it would re-narrow. I restructured the reader to return directly instead of routing through that optional:

```cpp
auto opt = getFloatVar(name);
if (opt.has_value()) return static_cast<T>(opt.value());
auto intOpt = getIntVar(name);
if (intOpt.has_value()) return static_cast<T>(intOpt.value());
throw ...
```

So an **integer** var read as a `double` is now converted `int64_t -> double` once, instead of being rounded to `float` precision first. `test_get_var_double_does_not_round_integers_through_float` pins it with `9007199254740991` (exact as a double, not as a float), which returned `9007199254775808` before.

One limitation this does not fix: a var declared as an actual **float** still goes through `std::optional<float>`, since the generated `getFloatVar` accessor is typed that way — so `getVar<double>` of a float var remains float-precision. Widening that accessor to `double` is a slightly bigger change (`header.py`'s `_VarType` table, the generated signature, and the tests asserting it). Happy to do it here or in a follow-up — let me know which you'd prefer.

rsalesc: Correction to the number in my previous reply: the old float round-trip returned `9007199254740992`, not `9007199254775808` (I verified it). The point stands — off by one from the declared `9007199254740991` — I just quoted the wrong value.

docs/setters/variables.md
```diff
@@ -52,8 +52,15 @@ In this section, we go through the different ways to use variables in each of th
...
-`getVar<bool>(name)`, `getVar<int>(name)`, `getVar<float>(name)` and `getVar<std::string>(name)`.
+of a variable as a `T`-typed object. `T` can be `bool`, `std::string`, any
+floating-point type (`float`, `double`) or any integer type -- both the plain
+spellings (`int`, `long`, `long long` and their unsigned counterparts) and the
+fixed-width aliases (`int32_t`, `int64_t`, ...). Reading an integer that does not
+fit in the requested type raises an error instead of silently truncating it.
+
+Character types (`char` and friends) are deliberately not supported: it would be
+ambiguous whether `getVar<char>("x")` should give you the number or the letter.
+Read the variable into an explicit integer type instead.
```
rsalesc: Please remove everything after --

PR #637, opened 2026-08-12
feat(packaging): add a moj-next packager

rsalesc:
Adds `rbx package moj-next`, a **separate** packager targeting MOJ (`moj.naquadah.com.br`) as [`cd-moj/mojtools`](https://github.com/cd-moj/mojtools) actually consumes it. The existing `moj` packager is untouched and keeps serving interactive problems.

Statements are deliberately out of scope for this iteration — a dummy `docs/enunciado.md` is written instead.

## Why a new packager

Researching the format against mojtools showed the legacy packager targets a shape MOJ no longer accepts:

| Area | MOJ today | Legacy `moj` packager |
|---|---|---|
| Time limit | **Measured, never authored** — `calibreitor.sh` scales the worst `sols/good` time by `TLMOD[calibrafactor]` | Writes `tl` only under `--for-boca` |
| Checker | `scripts/checker.cpp` + a ~10-line **stub** delegating to mojtools' bridge | Ships a private copy of the bridge, plus `testlib.h`/`rbx.h` in `scripts/` |
| Memory | `MEMLIMITMB` (peak RSS) | `ULIMITS[-v]`, the legacy knob |
| Statement | `docs/enunciado.{md,org,tex}` | `docs/enunciado.pdf` — not a recognized format |
| Tests | Samples required, named `sample*` | Everything `001`/`002` → **no samples**, validation fails |
| Subtasks | `tests/score` | Not emitted |

Reshaping a `BocaPackager` subclass into something sharing almost nothing with BOCA was the wrong move, so this extends `BasePackager` directly — mirroring the `boca_next` precedent already in the tree.

## The two constraints that shaped the design

**Stub on the host, copy in the jail.** From `mojtools/CLAUDE.md`: what runs on the judge host ships as a stub pointing at mojtools; only what enters the sandbox is a real copy. A bundled bridge copy once replicated a `bwrap` bug into **198 packages**, where fixing mojtools reached none of them. So `scripts/compare.sh` is a byte-copy of mojtools' canonical stub, and only `scripts/<lang>/{compile,run}.sh` are real copies.

**The checker must be one file.** `checker-bridge.sh` compiles it under `bwrap` with only `checker.cpp` and `testlib.h` bound into `/tmp`. A checker including `rbx.h` or a local header would report a judge error on *every* test.

## What's here

- **`rbx/box/dependencies/amalgamation.py`** — a reusable, package-agnostic amalgamator reducing a source and its closure to one translation unit. Diamonds collapse, cycles terminate, `#pragma once` is dropped, `<system>` includes are untouched, and `extra_roots` lets a caller make builtin headers inlinable without the library knowing what they are. Built on a new `DependencyScanner.reference_spans` splice capability (`rewrite` only renames a path; splicing needs the whole directive's span). Reusable by Polygon upload or BOCA's heredocs.
- **`rbx/box/packaging/moj_next/`** — the packager. No `tl`; `conf` carries `MEMLIMITMB`, `ULIMITS[-f]` and `TLMOD[calibrafactor]=1.35`. Checker and every C/C++ solution amalgamated, with packaging **refused** (actionable message) when a closure can't be inlined.
- **Naming/scoring** — `sample001…` plus `t<NN>_<group>_<NNN>`, so samples sort first in MOJ's lexicographic judging loop and group order follows `problem.rbx.yml`. Both of `score-summary.sh`'s group-matching heuristics agree by construction. `tests/score` for POINTS only.
- **`.moj-meta.json`** — the server is two-tier on tar uploads: it sources `display_title`/`collections`/`languages` *from* the tar, and refuses the access fields. So rbx writes `display_title` and `languages`, where `languages` is derived from the languages with an accepted solution — MOJ's own criterion, since it calibrates a time limit per language from `sols/good` and a language without one can never be submitted in. `public`/`public_at`/`owner` are never written: ignored from a tar anyway, and `public` is fail-closed server-side. Because the whitelist is derived rather than authored, packaging **says so out loud**: it prints the enabled languages and names every emitted language left out for having no `ACCEPTED` solution, so nobody discovers a C++-only problem on the judge.
- **`MojLanguageExtension`** (key `moj`) mirroring `BocaLanguageExtension`, wired into the default preset for c/cpp/py/java/kt.
- **Java/Kotlin via manifest jar** — `compile.sh` names the entry point so `run.sh` is just `java -jar`. Deliberately better than MOJ's own `lang/java/compile.sh`, which elects the main class by grepping for `main` and falling back to `ls *.class` — locale-dependent once javac emits nested `Main$X.class`.

## Verification

- 65 new tests; `tests/rbx/box/packaging` + `tests/rbx/box/dependencies` = **292 passed** (the 3 errors are the docker-dependent BOCA e2e tests, pre-existing locally).
- Built a real package end to end with `rbx package moj-next` and inspected the output: correct layout, `sample001` sorting before `t01_main_00N`, solutions split across `good`/`slow`/`wrong`, every `.sh` at 0755 inside the zip, no `tl`, no `scripts/testlib.h`. Read the emitted `.moj-meta.json` back with the same `jq` expressions `gen-problem-json.sh` uses, confirming `display_title`/`languages` parse and `public` resolves fail-closed to false.
- **Ran the emitted `scripts/compare.sh` through the real mojtools bridge**: correct output → exit 4 (Accepted), wrong output → exit 6 (Wrong Answer).
- The amalgamated checker (~207 KB with testlib inlined) compiles at `-std=gnu++17` with MOJ's forced `-include cassert -include cstring -include cstdint`.

Full-suite failures on this branch are the machine's known pre-existing ones (C++ compiled with clang rather than g++, plus docker); verified identical with the shared-file changes reverted to base.

## Known gaps

- **The package is unjudgeable until a judge calibrates it**, and rbx's authored `timeLimit` is advisory. mojtools' README says the agent calibrates on first submission, but that logic lives in the private `judge` repo — worth confirming with MOJ operators before the first real upload. There's a `TODO` in `_write_conf` to derive `calibrafactor` from `timeLimit / measured model-solution runtime`.
- **Per-test partial credit is unavailable** — testlib `quitp`/`_points` maps to a judge error on MOJ. The packager warns when the checker source uses them.
- **Interactive is out of scope** — MOJ's arbiter protocol is structurally unlike a testlib interactor and deserves its own design.

Design doc: `docs/plans/2026-08-13-moj-next-packager-design.md` · Plan: `docs/plans/2026-08-13-moj-next-packager.md`


rbx/box/packaging/moj_next/packager.py
```diff
@@ -0,0 +1,487 @@
...
+    def statement_types(self) -> List[StatementType]:
+        # MOJ's statement format is not supported by rbx yet; a dummy enunciado.md is
+        # written instead, so no statement is built.
+        return []
+
+    # -- metadata -------------------------------------------------------------
+
+    def _display_title(self) -> str:
+        pkg = package.find_problem_package_or_die()
+        for language in _TITLE_LANGUAGE_PREFERENCE:
```
rsalesc: i think there are more reliable ways of getting the problem name, see the BOCA packager, using the naming module.

rsalesc: Good call — switched to `naming.get_problem_title(language, statement, fallback_to_title=True)`, the same helper `BocaPackager._get_problem_info` uses, and dropped my packager-local rule (which preferred `pt`/`pt-br`/`en` and then the lowest language code).

Besides being the shared path, it is strictly better here:

- it honors a statement's own `title` override, which my version ignored entirely;
- it takes the language from the main statement instead of a hardcoded preference list;
- on an ambiguous title (several `titles`, no statement) it raises the actionable error rather than silently picking one.

Added `_get_main_statement()` returning the topmost declared statement — statements are not *built* for MOJ (`statement_types()` is `[]`), but a package still declares them, so the main-statement path works normally.

Tests updated: `test_display_title_uses_the_package_title`, `test_display_title_reports_an_ambiguous_title` (documents the shared error behavior), plus the existing fallback-to-`pkg.name` case.

In 06cdcecd.

PR #666, opened 2026-08-18
feat(vscode): surface a solution's expected outcome in the run view

rsalesc:
## What

Follow-up to #664. That gave every verdict its own icon, settling how the Run view shows **what happened**. This settles how it shows **what the package asked for**, and whether it got it.

Two problems, both visible in the same row:

1. `sols/wa.cpp`, declared `WRONG_ANSWER`, answers wrongly and draws a red `close` icon — pixel-identical to the main solution breaking. The declaration was only in the hover.
2. A solution that *missed* its declaration said `expected AC, got WA` in dim grey `description` text, sitting among every other row's `AC · 120 ms · 4 MiB`. Nothing drew the eye.

## How

`rbx run` already splits these onto two axes: `_render_detailed_group_table` heads each solution's **column** with `solution.href()`, coloured by `ExpectedOutcome.full_style()`, and fills the **cells** with what happened. A tree row has one line, so the split moves to a channel a `TreeItem` actually has.

**The icon keeps reporting the outcome** (#664 untouched, so a TLE stays yellow in the tree exactly as it is yellow in the terminal beside it). **A `FileDecoration` reports the expectation:**

| | |
|---|---|
| **badge** | the declared expectation, as `ExpectedOutcome.icon()` marks it: `✓` / `⧖` / `✗`. `ANY` gets none — nothing was declared. |
| **colour** | the match axis only: `charts.red` on a miss, nothing otherwise. |

Colouring only misses is a deliberate divergence from the CLI, which tints the expectation by its own hue. One channel cannot carry both facts, and colouring by what was *declared* leaves a mismatch with no colour of its own — which is problem 2 unsolved. Because nothing else is coloured, misses are the only coloured rows.

`FileDecoration.badge` is plain text and cannot hold a codicon, so badge and icon are necessarily different shapes. They are picked to rhyme: `pass` is a tick in a circle beside a `✓`, `close` a cross beside `✗`.

Solutions decorate by their **real file URI**, so the declaration also rides along in the Explorer and on editor tabs — you see what a solution promises while editing it. Groups get a synthetic `rbx-run:` URI and are badged only where `outcomePerGroup` covers them; testcases get nothing, having no expectation of their own.

## Staleness

The badge comes from `problem.rbx.yml` and can never go stale. The colour is run-derived, and on an editor tab it would go on accusing a file the user has already fixed — so a solution newer than `report.yml` keeps its badge and loses its colour.

## Consolidating the two vocabularies

`outcome.ts` opens by warning that `Outcome` and `ExpectedOutcome` are different types that must not be conflated, and then ended with:

```ts
default:
  return shortName(expected.toLowerCase().replace(/_/g, '-'));
```

That reaches an `Outcome` key by lowercasing an `ExpectedOutcome` member name. It holds today by coincidence, not contract — the enums have genuinely different members (`INCORRECT`, `ANY`, `TLE_OR_RTE` exist only in one; `SKIPPED` only in the other). Hanging a badge off it is what would have made the two vocabularies visibly disagree.

Now: two explicit tables sharing one glyph alphabet, no string-guessing between them. `DISPLAY` gains a `glyph` field mirroring `get_outcome_markup_verdict`; a new `EXPECTED` table has one record per `ExpectedOutcome` member, with a test asserting every member is covered.

The glyph is **stored, not derived from the outcomes an expectation matches**, because rbx does not derive it either — `ACCEPTED_OR_TLE` matches both `ACCEPTED` and TLE yet is marked `✓`, and `INCORRECT` matches five outcomes yet is marked `✗`. Any derivation gets both wrong.

## Also

- Pending rows say what they promise: `expects AC · 3/10` rather than `3/10`. The declaration is in the skeleton, so it is known before and during the run.
- A missed solution gets `rbx.solution.mismatch` as its `contextValue` (suffixed, so the existing prefix-regex menu `when` clauses keep applying).
- The view carries a badge counting misses, visible with the view collapsed. It counts *misses*, not failures — badging a solution that failed on purpose would report the package's own test suite as a problem.
- `createTreeView` replaces `registerTreeDataProvider`; the handle is the only way to set that badge.

## Testing

`npm run typecheck`, `npm run compile`, `npm test` — **42 pass, 0 fail** (up from 25; 17 new). `eslint` is not installed in this checkout and `npm run lint` fails on `main` too.

Verified end to end against a real `rbx run` of `tests/e2e/testdata/outcome-per-group`, which exercises met, per-group-missed and pooled cases at once:

```
--- sols/main.cpp        badge ✓  colour —          Expected ✓ AC, got ✓ AC
--- sols/partial.cpp     badge ✗  colour —          Expected ✗ INCORRECT, got ✗ WA
      group small ✓      group big ✗
--- sols/mislabeled.cpp  badge ✗  colour charts.red Declared ✗ INCORRECT, but small, big did not match
      group small ⧖ red  group big ⧖ red
view badge count: 1
```

`partial.cpp` is the point of the whole change: it fails (red `close` icon, WA) but is doing exactly what the package declared, so its row stays calm. `mislabeled.cpp` is the only red row.

That run is also what caught a real bug in the first draft of the hover: it read `Expected ✗ INCORRECT, but got ✗ WA`, yet `mislabeled.cpp`'s pooled `INCORRECT` was **met** — it does fail — and only the per-group layer caught it. That is the trap `summary.ts` already documents ("would name an expectation that was in fact met"). The hover now names the failing groups instead, with a regression test.

Staleness was checked by touching a solution and confirming the colour drops while the badge stays.

Design doc: `docs/plans/2026-08-18-vscode-expected-outcome-design.md`.

## Notes

No rbx-side change and no report v2: `report.yml` v1 already carries `expectedOutcome`, `matchesExpectation` and `failedGroups` per solution and per group, and `skeleton.yml` carries the declaration before any run finishes. This is presentation only.


rsalesc:
## Update: the icon is now the expectation

Two rounds of review changed the shape of this PR. Recording where it landed.

**The complaint:** the row drew a codicon on the left and one of rbx's `✓ ⧖ ✗` glyphs on the right — two alphabets for the same kind of statement.

**The fix:** the glyphs are gone; the tree speaks codicons. And the split moved.

A solution row is a *claim the package makes*, not a result — which is what `rbx run` already says by heading each solution's column with `solution.href()` coloured by `ExpectedOutcome.full_style()`, and putting verdicts in the cells below. So:

| channel | carries |
|---|---|
| **icon** | the declared expectation, one per `ExpectedOutcome` |
| **description** | what the run actually produced |
| **label colour** | red **iff** the run missed the declaration |
| **badge (right)** | `✗` **iff** the run missed the declaration |
| **tooltip** | full breakdown, naming every group that missed |

Testcase rows are untouched and keep #664's verdict icons — which is also what keeps the two sets rhyming: an `ACCEPTED` row and an accepted testcase beneath it both draw `pass`.

### One icon per expectation

Where an expectation names exactly one outcome it borrows that outcome's icon. The four that name no single outcome get their own, so the pairs a coarse pass/fail/slow set would have collapsed stay distinct:

| expectation | icon | | expectation | icon |
|---|---|---|---|---|
| `ANY` | `dash` | | `TIME_LIMIT_EXCEEDED` | `watch` |
| `ACCEPTED` | `pass` | | `TLE_OR_RTE` | `flame` |
| `ACCEPTED_OR_TLE` | `pass-filled` | | `MEMORY_LIMIT_EXCEEDED` | `server` |
| `WRONG_ANSWER` | `close` | | `OUTPUT_LIMIT_EXCEEDED` | `arrow-both` |
| `INCORRECT` | `circle-slash` | | `JUDGE_FAILED` | `law` |
| `RUNTIME_ERROR` | `zap` | | `COMPILATION_ERROR` | `tools` |

A test holds all twelve distinct, and every id was checked against the codicon table in an installed VS Code build. Colours are `ExpectedOutcome.style()` transposed — deliberately *not* the verdict palette: a declared OLE is magenta while an OLE verdict is orange, and copying one onto the other would invent a colour rbx never prints.

### Red means exactly one thing

Only a missed expectation colours a label. Verdict colour lives in the icon, so the label channel carries one fact. Descriptions no longer repeat the expectation now that the icon shows it:

| case | description |
|---|---|
| met | `AC · [100/100 pts] · 12 ms · 10 MiB` |
| missed, pooled | `got WA · [30/100 pts] · 9 ms · 10 MiB` |
| missed, per group | `failed samples, main, edge, big · ...` |
| still running | `expects AC · 3/10` |

The per-group form matters: `mislabeled-groups.cpp` declares `incorrect` pooled — which **holds**, it does fail — and `tle` per group, which it misses. Naming the pooled declaration there would accuse an expectation that was met.

### Rows are no longer files

Every decorated row now uses a synthetic `rbx-run:` URI, **solutions included**. Decorations are global per URI, so pointing at `sols/wa.cpp` marked it in the Explorer and on its editor tab too. That is a separate question with its own scoping and staleness problems — deferred to #669. Nothing else reads `resourceUri` (the commands all take the node), so nothing else changes.

This also retires the mtime-vs-report staleness rule from the earlier draft: its whole purpose was to stop an editor tab accusing a file the user had already fixed, and no decoration leaves the Run view any more.

Also drops the custom mismatch icon font a previous attempt added — it existed to composite a mismatch mark onto a verdict icon, and the mark is now the badge. That removes a 247-line font generator, a checked-in `.woff`, and 86 lines of `contributes.icons`.

## Verification

`npm run typecheck`, `npm run compile`, `npm test` → **42 pass, 0 fail**.

Verified end to end against `rbx-vscode-sample`, which gained three solutions covering all three mismatch shapes (pooled, pooled-reverse, per-group-only). Rendering the real `report.yml` through the shipped code:

```
       pass          green   sols/main.cpp               AC · [100/100 pts] · 12 ms
       pass-filled   green   sols/slow-ok.cpp            AC · [100/100 pts] · 143 ms
       circle-slash  red     sols/partial.cpp            WA · [30/100 pts] · 10 ms
       close         red     sols/wa.cpp                 WA · [0/100 pts] · 10 ms
       watch         yellow  sols/tle.cpp                TLE · [0/100 pts] · 2007 ms
       watch         yellow  sols/ile.cpp                ILE · [0/100 pts] · 2000 ms
       zap           blue    sols/rte.cpp                RTE · [0/100 pts] · 20 ms
       server        yellow  sols/mle.cpp                MLE · [0/100 pts] · 93 ms
✗ RED  pass          green   sols/optimistic.cpp         got WA · [30/100 pts]
✗ RED  close         red     sols/too-pessimistic.cpp    got AC · [100/100 pts]
✗ RED  circle-slash  red     sols/mislabeled-groups.cpp  failed samples, main, edge, big
     ✗ RED  watch              samples                   got AC · 8 ms
     ✗ RED  watch              main                      got AC · [30/30 pts]
     ✗ RED  watch              edge                      got WA · [0/30 pts]
     ✗ RED  watch              big                       got WA · [0/40 pts]

view badge: 3
```

Eight healthy rows, no badges, no red — six of them drawing a "failure" expectation while doing exactly what the package asked. Three real problems, unmissable.

## Known limitation

Hover *delay* is not controllable from an extension — it is the user-level `workbench.hover.delay` setting (default 500ms). The tooltips are richer, but making them appear faster is a settings change, not a code one.

PR #673, opened 2026-08-19
fix(vscode): tell the two declaration layers apart in the run view

rsalesc:
## What

Follow-up to #672, from the `mislabeled-groups` solution in the sample package.

That solution declares `outcome: incorrect` with `outcomePerGroup: {'*': tle}`. It **is** incorrect — it fails on `edge` and `big` — so the pooled layer *holds*. It is caught only by the per-group layer, which every group missed. The view said of it:

> ⚠ Declared INCORRECT, but samples, main, edge, big did not match.

Both halves are true and the sentence is false: those groups missed `TLE`, not `INCORRECT`. And nothing anywhere in the view said `TLE` at all — the four group rows were hueless names with a warning beside them, so the reader could see *that* something was wrong and had no way to see *what was wanted*.

Two bugs with one root: the run view models one declaration where a solution can carry two.

### The report says which layer failed

`matchesExpectation` is the aggregate of the pooled `outcome` and the per-group `outcomePerGroup`, and the aggregate cannot say which one to blame. rbx already draws this distinction — `get_verdict_markup` will only print `Expected: X` when `pooledStatus` is itself not ok, precisely so it does not accuse an expectation that was met — but `pooledStatus` was the one fact the published report never carried, so the extension had no way to reach the same conclusion.

`RunSolutionReport` gains `pooledMatchesExpectation`. `expectedScore` comes with it, so the third failure mode — a score outside its declared range — has something to say beyond "wrong".

Both fields are optional, so **the version does not bump**: an older reader drops what it does not know and reads the rest unchanged, where a bump would instead make every existing reader ignore the whole file. An extension meeting a report written before this change infers the pooled layer from "no group failed, so nothing else can have" — exact except when both layers failed at once, where it under-reports rather than accusing a layer that held.

### A row spells its declaration

The expectation gets a channel of its own, between the meta and the verdict, on solutions and groups alike:

```
[gutter][twisty][label .............][meta][expectation][verdict]

  ⚠  edge      [0/30 pts] · 9 ms · 10 MiB    TLE → ✗ WA
```

It grows leftwards from the verdict rather than taking a fixed column of its own. The verdict column is fixed-width so it can be scanned straight down; giving this one the same treatment would mean reserving the width of `TLE or RTE` on every row in a sidebar this narrow. What has to line up is the *pair*, and the pair is anchored on the right where the arrow meets the verdict.

Groups take the label hue too, by the rule solutions already used — #672 gave groups the gutter but left them out of the expectation channel, which is the gap. `ANY` still draws nothing: it is how a setter declares nothing, and the gutter already treated it that way. Under a container query the declaration drops after the timings and before the verdict, which is the order of how much each would be missed.

### The card speaks per layer

`MismatchDetail` becomes a set of optional clauses — a pooled `declared → got` pair, a list of groups each carrying its own pair, a score range — and the renderer prints only the clauses that are set. No sentence it can produce accuses an expectation that held.

`mislabeled-groups.cpp` now reads:

```
⚠ INCORRECT held for the solution as a whole; 4 groups missed their
  outcomePerGroup declaration:
     samples   TLE → AC
     main      TLE → AC
     edge      TLE → WA
     big       TLE → WA
```

Naming the pooled layer as **held** is deliberate rather than redundant: the reader is looking at a row labelled `INCORRECT` and needs to be told that is not the declaration these groups missed.

`optimistic.cpp`, which misses only the pooled layer, reads `Declared AC → WA` and says nothing about groups. `partial.cpp`, which meets both layers while failing on purpose, stays completely calm as before.

The card is no longer red throughout — the icon carries the alarm. That was right when it held one red sentence; now that each pair is hued by what it actually is, a red wash sat in front of the colours doing the work.

## Verification

Checked against the real `rbx-vscode-sample` run rather than only fixtures: regenerated its report with the patched rbx (`pooledMatchesExpectation: true` for `mislabeled-groups`, `false` for `optimistic`), then rendered it through `viewModel` + `render` and looked at the result in a browser with VS Code's dark variables filled in.

- `npm run typecheck`, `npm run compile`, `npm run package` clean; `npm test` **84 pass, 0 fail** (76 → 84).
- `uv run pytest tests/rbx/box/run_report_test.py` 10 pass; `run_report_test.py` + `solutions_test.py` 85 pass.
- Full `tests/rbx` (minus CLI): 35 failures, byte-identical to the same run on the base commit — the pre-existing local C++/sandbox set. 4550 → 4552 passed, which is the two new tests.
- Layout measured in-browser: no row overflows at any width, verdict column stays aligned, and the degradation ladder fires as designed — meta drops at 260px, expectation at 200px, verdict always survives.

Tests pin the traps rather than the markup. `viewModel.test.ts` asserts `pooled === undefined` for the group-only miss (naming the pooled layer there is the whole bug), asserts the stale-report inference separately, and asserts a group inherits no declaration from the solution above it. `render.test.ts` asserts the card never emits `Declared <INCORRECT>` for that solution, and that a group named `` reaches the output escaped.

## Notes

Design doc updated: the two superseded paragraphs in D3/D4 are marked as such rather than rewritten, and D9 records the correction. The `summary.ts` logic D4 said it would reuse never made it into the implementation — that intent is now enforced by the model shape instead of by prose.

### DIFF
--- a/docs/plans/2026-08-19-vscode-run-webview-design.md
+++ b/docs/plans/2026-08-19-vscode-run-webview-design.md
@@ -227,8 +227,8 @@ a nonce CSP with no remote origins.
 
 - **Accessibility is now ours.** Screen-reader quality depends on D7 being done
   properly rather than on VS Code doing it for us.
-- **Width.** At roughly 300px the meta column has to drop out under a container
-  query, and the detail card stays single-column.
+- **Width.** The row sheds channels in priority order as the sidebar narrows
+  (D10); the detail card stays single-column.
 - **Virtualization is deferred.** Rendered rows are solutions plus groups, plus
   testcases only under an expanded solution — and only a solo run expands one by
   default. Revisit past ~2000 rendered rows.
@@ -277,8 +277,7 @@ is what has to be read together, and reserving the width of `TLE or RTE` on
 every row costs more than a ragged left edge does. Groups take the label hue
 too, by the same rule solutions already used. `ANY` still draws nothing --
 it is how a setter declares nothing, and the gutter already treats it that way.
-Under a container query the declaration drops after the timings and before the
-verdict, which is the order of how much each is missed.
+Under a container query the declaration drops **first** -- see D10.
 
 **The card speaks per layer.** `MismatchDetail` becomes a set of optional
 clauses -- a pooled `declared → got` pair, a list of groups each with its own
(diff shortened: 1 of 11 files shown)


rsalesc:
## Update: the responsive ladder (second commit)

The declaration channel added above is genuinely more to fit on a row that was already tight, so this makes the narrowing behaviour explicit and ordered instead of leaving it to two ad-hoc breakpoints.

The first cut hid the meta line **whole** at 260px and the declaration at 200px. Both halves were wrong: it took the score away along with the memory figure — one blob, one switch — and it kept the *declaration* alive longer than the measurements, when the declaration is the least urgent thing on a cramped row.

The order is now by how much each would be missed, not by how much room each frees:

| # | dropped | at | why it goes when it does |
|---|---|---|---|
| 1 | the declaration | 400px | the gutter still answers *was it met*, which is the part needed at a glance; the card says what it was |
| 2 | memory | 330px | the measurement least often being looked for |
| 3 | time | 280px | wanted often, but a solution is opened to study its timings, and the card carries the maxima |
| 4a | the verdict's **name** | 240px | the colour and shape carry most of what the name says, at a tenth of the width |
| 4b | the verdict's icon | 190px | below this a row is a path and a score — the least it can be and still be worth drawing |
| 5 | the score | never | on a points problem it is the answer to the question the view exists to answer, and the cheapest thing on the row |

Everything above the score is recoverable by widening the sidebar or opening the row. The breakpoints sit where the label — a path, the one thing recoverable from nowhere else — would otherwise be squeezed past legibility.

### Two mechanics

**Spans are named.** `Span` gains a `role` (`progress`/`score`/`time`/`memory`) which the renderer turns into a `span-*` class. An anonymous list of strings can only be shown or hidden whole, which is exactly what forced the original all-or-nothing switch.

**Separators belong to the span that follows them.** `metaCell` now emits nothing between spans; the stylesheet draws the `·` as a `::before` on every span but the first. A separator written as its own element would outlive the span it divides and leave `[30/100] ·` trailing off a narrowed row.

That is correct only while hiding removes a **suffix** of the line — so the meta line is built in priority order, and a `render.test.ts` case asserts the breakpoints are strictly descending and that the score is named by none of them. Reorder the ladder and that test fails, rather than the separators quietly going wrong.

### `pts` dropped

`formatScore` now returns `[70/100]`. In a terminal the unit earns its place; in a sidebar it is four characters of the one span that has to survive everything else, and the brackets already say what the number is.

### Verification

Measured in-browser against the real sample run at 480 / 420 / 380 / 320 / 270 / 230 / 185 / 170px:

- the ladder fires in exactly the order above, and **no row overflows at any width**;
- the label keeps ≥94px down to 230px, degrading only below that;
- separators confirmed via `::before` computed content — the score never carries one, and time and memory each take theirs with them when they go, so a dangling `·` is not reachable.

`npm run typecheck` / `compile` / `package` clean, `npm test` **86 pass, 0 fail** (84 → 86).

Design doc: D10 records the ladder and both mechanics; D9's now-reversed sentence about the declaration outliving the timings is corrected, and the stale "at roughly 300px the meta column drops out" risk bullet points at D10.

PR #697, opened 2026-08-23
feat(run): draw a solution as one live block with a wall clock

rsalesc:
Follow-up to #692. First of two: this one is backend-agnostic and useful on its own.

## Why

`LiveRunReporter` started a `rich.live.Live` per **group**, and printed the solution header ahead of the first one. So by the time anything interesting was known about a solution, its header was already frozen in scrollback — the wrong scope for the thing a setter actually waits on. During a remote run (`rbx time --runner moj`) rbx can sit for minutes with nothing moving and no way to say why.

## What

The live region moves up to the **solution**. The header and the group lines indented under it animate as one block and finalize together, so the header has somewhere to carry state that only becomes true *while* the solution runs.

```
sols/slow.cpp (runs/2)  · 12.4s
  samples (3)  (120ms, 4MB)
  main (12)    4/WA (890ms, 12MB)
  big (7)      3/.. 5/..  (1.1s, 12MB)
```

The first such chip is a wall clock, so a slow solution and a hung one stop looking identical. The second — what the backend running it wants said (testrun id, judge, queue state, cache hits) — is the follow-up PR.

Group-line rendering is unchanged: `_group_line` is the old `_update_live` body, moved verbatim. What each line *says* is byte-identical; only where it sits moved.

## Keeping this out of recorded output

- **Wall-clock chips render on a terminal only.** `--share` reports, e2e goldens and asciinema casts are all non-terminal consoles, and an elapsed time in any of them is a diff on every run.
- **Auto-refresh is terminal-only** too. The clock has to repaint with nothing else happening, but a refresh thread on a non-terminal would spin for frames `rich` never emits (`Live.refresh` is a no-op there until `stop()`).

## Height guard

`rich` redraws a live region by moving the cursor back over its own output, so a block taller than the terminal flickers or tears. `_fits_as_block` falls back to the per-group Live when the groups do not fit — which is exactly the behaviour that shipped before this block existed.

## Drive-by fix

The report loop had no `finally`. A deferred that raises — the normal shape of a judge that never answered — unwound with the Live still started, hiding the cursor and eating the first lines of the traceback. `TraditionalRunReporter.close()` now runs from a `finally` around the loop.

`_Elapsed` moves from the MOJ runner to `utils.Elapsed`, so a header and a `prepare` summary line cannot disagree about how long something took.

## Tests

New: block ordering, indentation, terminal-only clock, both sides of the height guard, Live teardown on a raising deferred, `close()` idempotence, `Elapsed` formatting.

`tests/rbx/box/solutions_test.py` — `rendered_group_lines` now accepts the block indent, deliberately *not* by stripping first: a multi-group failure message continues on indented lines that also name a group.

Verified: `solutions_test.py` (84), plus `test_sharing.py`, `test_run_cli.py`, `test_timing*.py`, `run_report_test.py` (106) — all passing.

Design: `docs/plans/2026-08-22-live-run-progress-design.md`.

https://claude.ai/code/session_01Jj1dDYMUJByG2yxLT1JQhq


rsalesc:
Pushed `6f5c0039` — a real bug found in testing this branch.

**Symptom:** on `rbx time --runner moj` the header sat on `waiting for a slot` with the clock frozen at `0.0s` for the whole judge wait, then jumped straight to finished.

**Cause:** `rich.live.Live` redraws the renderable it was *handed* — `Live.refresh` re-renders `self.renderable`, the object passed to `update()`. This branch assembled the block as a `Text` with the clock already formatted into it, so every auto-refresh redrew an identical frame. The display only changed when something called `update()` again, which here means an evaluation resolving — and on a remote run the first one does not resolve until the judge has finished the entire testrun. The whole span worth watching is the span in which nothing calls `update()`.

**Fix:** hand `Live` a `_SolutionBlock` that renders the block afresh each time it is asked. The refresh thread then produces frames showing what is true now, without the reporter needing to know when the interesting changes happen — which it cannot know, because on a remote backend they happen on someone else's task.

`block_renderable` consequently runs on the refresh thread too. It only reads, copies the finished-line list before walking it, and the values it reads are immutable, so a write landing mid-frame cannot be seen half-applied.

Two regression tests, both of which fail against the previous commit.

PR #698, opened 2026-08-23
feat(runners): let a backend say what it is doing on the solution header

rsalesc:
Stacked on #697 — review that one first; this PR's diff is against it.

Second of two. #697 gave the solution header a live region; this fills it with what the backend running the solution actually knows.

## The problem

`MojRunner.run_solution` dispatches **every** solution onto a background task immediately, so while the report is blocked on solution #1, solutions #2..#N are genuinely in flight on the judge. rbx knows their testrun ids, their queue state, whether a result came from cache — and had nowhere to say any of it.

It was not for lack of trying. `_evaluation_from_job` called `ctx.progress.update('Waiting for MOJ to judge …')`, but **every caller exits its `with utils.StatusProgress(...)` block before `print_run_report`** (`cli.py`, `timing.py`). The `Status` was always already stopped, so that line rendered nowhere. It has never been visible to a setter.

## The board

`RunProgress` — one slot per solution, holding `RunnerChip`s (text + style). Created by `run_solutions`, handed to the backend on `RunContext` and to the reporter on `RunSolutionResult`. The two ends cannot be wired directly: the reporter does not exist until `run_solutions` has returned.

```
sols/slow.cpp (runs/2) · 12.4s · moj alice#rbxt-a1b2 · testrun 4821 · running · judge-sp1 · 34/72 ok · 51s
```

**Chips, not a typed status.** A typed `RunnerStatus` would put MOJ's vocabulary — `queued`, `running`, `done` — into the reporter, and the next judge's after that. The backend owns its words; the reporter owns the layout.

**Pull, not push.** The reporter reads the slot it is blocked on, each time it paints.

This dissolves the standing objection to a backend saying anything from a polling task. `_submit_and_poll`'s docstring argued that a poll writing a status line "makes the display flip between solutions and the reporter's own message every few seconds" — true of a *print*, false of a board write, which lands in that solution's own slot and is drawn only if the reporter is waiting on it. So all ten in-flight solutions keep their slots current for a dict write each, and the one being watched reports itself live. Those docstrings are rewritten rather than bypassed.

Same event loop on both ends, so no locking; chips come back as a tuple so a reader holds a snapshot. `LocalRunner` writes nothing and reads back empty.

## What MOJ writes

| Where | Chips |
|---|---|
| `run_solution`, at dispatch | `moj <id>`, `waiting for a slot` |
| `_submit_and_poll`, cache hit | `testrun <id>`, `cached` |
| `_submit_and_poll`, after submit | `testrun <id>`, `submitted`, elapsed |
| `_wait_for_testrun`, each poll | `testrun <id>`, state, host, `N/M ok`, elapsed |
| `_submit_and_poll`, on done | `testrun <id>`, `done`, elapsed |
| failure | nothing — the exception speaks |

Two judgement calls worth checking:

- **Only what the judge answered with.** The state word is always present; counts and host are not, so they are added when reported rather than defaulted. A `0/0` on a run that has not started would read as a judge losing the testset.
- **`correct`, never "tests run".** MOJ reports how many *passed*. Calling that progress would show a solution the validation phase expects to fail sitting at `0/72` while working perfectly — so the chip is labelled `ok`.

The elapsed chip starts **before** the concurrency slot is taken, so it counts the whole wait the setter is enduring. On `MAX_INFLIGHT_TESTRUNS = 1` that queueing is most of it, and it is the part the reporter's own clock cannot see — that clock starts when the *report* reaches the solution, long after the judge did.

The slot is cleared once the solution has a real verdict: a stale `running` chip beside a finished outcome is worse than no chip.

`_evaluation_from_job` deliberately writes **nothing** on the way in. The obvious move — replacing the dead `progress.update` with a `waiting for the judge` chip — is wrong: that solution's own polling task is already repainting the slot with the testrun id, the judge's state and the elapsed wait, so a write here would overwrite a better line with a worse one and leave it worse until the next poll came round.

`TestrunStatus` gains `host` (the 2026-08-21 probe saw `judge-sp1`), so a timing can be read against the machine that produced it.

## Tests

- `tests/rbx/box/runners/test_progress_board.py` (new): whole-line writes, snapshot reads, per-solution slots, clear semantics, empty default.
- `tests/rbx/box/solutions_test.py`: chips reach the header; they survive a non-terminal console (unlike the clock); only the drawn solution's slot renders; an empty board adds no separator.
- `tests/rbx/box/runners/moj/test_run_solution.py`: each write point, plus `_poll_chips` saying only what the judge reported, plus `host` parsing.
- `test_the_background_polling_never_writes_the_status_line` keeps its assertion and gains a docstring saying why it still holds — it is about the *console*, and a board write is not a print.

https://claude.ai/code/session_01Jj1dDYMUJByG2yxLT1JQhq

### DIFF
--- a/rbx/box/CLAUDE.md
+++ b/rbx/box/CLAUDE.md
@@ -92,6 +92,24 @@ flattened testset in group order, and returns one `Deferred[Evaluation]` per ent
 grain is deliberate -- a remote judge judges one submission against every test at once, so
 a per-testcase seam would force every batch backend to coalesce calls back into a batch.
 
+**Saying what it is doing.** `RunContext.progress_board` is a `RunProgress`: one slot per
+solution, holding the chips a backend wants on that solution's header while the report waits
+on it. The same object reaches the reporter as `RunSolutionResult.progress_board`, because
+the two ends cannot be wired directly -- the reporter does not exist until `run_solutions`
+has returned. Chips are text plus a style, never a typed state, so the backend owns its own
+vocabulary and the reporter knows nothing about judges.
+
+The board is what makes a **background task** able to say anything. The standing rule was
+that a polling task must stay silent: the console and the `StatusProgress` belong to the
+reporter, and a poll printing from one of ten in-flight solutions makes the display flip
+every few seconds. A board write is not a print -- it lands in that solution's own slot, and
+the reporter draws only the slot it is currently blocked on. So every in-flight solution
+keeps its slot current for the cost of a dict write, while the one being watched reports
+itself live. Reads and writes share one event loop, so there is no lock; chips come back as
+a tuple, so a reader holds a snapshot. `LocalRunner` writes nothing and reads back empty.
+Anything that must be said *durably* still goes through the console from
+`_evaluation_from_job`. The board is a status line -- overwritten, then gone.
+
 Two things stay with the orchestrator rather than the backend:
 
 - **Abort/skip policy.** `_gated_evaluation()` wraps each returned deferred when a run asks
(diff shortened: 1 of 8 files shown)


rsalesc:
Pushed `a2d159d6` + `da42041e`, following the frozen-frame fix in #697 (see the note there).

**The board no longer clears the slot when a solution finishes.** Clearing was meant to stop a stale `running` chip sitting beside a finished verdict. Nothing was ever stale — `_submit_and_poll` overwrites the slot with `done` and the total wait, or with `cached`, before any deferred can resolve.

And once the block is rebuilt per frame, clearing actively loses information: the last frame — the one `Live.stop()` freezes into scrollback, and the *only* one a non-terminal console emits at all — renders whatever the board holds then. So clearing dropped the testrun id out of the permanent record and out of every `--share` report, which is where it is most worth having.

A finished solution's header now reads `moj alice#rbxt-a1b2 · testrun 4821 · done · 51s`.

`test_the_board_is_cleared_once_the_solution_has_a_verdict` is replaced by `test_a_finished_solution_keeps_the_testrun_on_its_header`, which also asserts nothing from the in-flight states survives.

**One test was worth less than it looked.** The end-to-end repaint test originally asserted on the finished transcript, and still passed when I reverted the fix to check it — the frame drawn *after* the deferred resolves carries the chip either way. It now snapshots the screen from inside the still-unresolved deferred, which is the only place that can tell "reached the screen during the wait" from "reached it once the wait was over". Against the bug it fails with the symptom as reported: `sol.cpp (runs/0) · 0.0s · waiting for a slot`.

Full suite re-run and diffed by exact test name against base `de1558e3`: **identical 45 failures** (the known local C++/sandbox and e2e ones), 5130 passed vs 5100.

PR #722, opened 2026-08-24
docs(profiling): restructure the profiling guide into six pages

rsalesc:
Restructures **Feature Guide → Profiling** from one 649-line page into six, on a single running example, and adds the two cast-pipeline knobs the recordings needed.

Design and plan: `docs/plans/2026-08-24-profiling-docs-restructure-design.md`, `docs/plans/2026-08-24-profiling-docs-restructure.md`.

## Why

The page interleaved three separate concerns — the `rbx time` flow, the rules that turn timings into a number, and profiles as a place to keep limits. It had no running example, so every snippet was a different toy problem, and its optional flags were sprinkled mid-narrative rather than sectioned. A reader who arrived because BOCA demanded a `boca` profile had to read the estimation theory to find out what a profile is.

The two behaviours readers most need to *see* — the upper-bound check catching a violation, and the language-group picker — were prose-only.

## The split

| Page | Answers |
| --- | --- |
| `index.md` | What profiling is, and the three-command happy path |
| `profiles.md` | What a profile is, and every command that reads one |
| `estimating.md` | `rbx time` in full |
| `computing.md` | Ratios, formulas, wall time |
| `language-groups.md` | Giving each language a limit that suits it |
| `remote.md` | Measuring on the judge park |

`profiling/remote.md` and `running/remote.md` are now siblings sharing `docs/_partials/moj-backend.md`, so the MOJ contract is written once instead of twice. This needed `base_path` on `pymdownx.snippets` — without it an include renders as literal text and says nothing about it.

## Recordings

One cast became four, on a new `timing-problem` fixture (count pairs summing to K: an accepted C++ and Python solution, a genuinely too-slow quadratic one, and Java left unsolved).

`time-estimate` went from **49.8s to 11.4s** while covering *more* than before — the old recording never showed the upper-bound check at all, because `ab-problem` has no solution it could catch.

The fixture carries its own environment in `.local.rbx/`. The ratios, the upper-bound check and the language groups all live in `env.rbx.yml` and nowhere else, and a problem cannot fill that gap on its own: `timing.multipliers` in `problem.rbx.yml` only *overrides* an environment that already sets them (`rbx/box/timing_config.py:76`).

## Cast pipeline

Two knobs, TDD, no behaviour change to existing casts:

- **`speed:` on an instruction, and `x8` tokens inside an `!Interactive` keys list.** Scales measured time only, so the typing animation keeps its authored pace. The mid-command form exists because `rbx time` holds both the seconds that let a reader take in a prompt and the minute that waits out the compute. Since key dwells are authored durations, a scaled window has the same length on any machine.
- **Per-instruction `width`/`height`**, emitting asciicast `r` resize events. Each instruction already gets its own pty, so this is a real terminal rather than a post-hoc crop, and `COLUMNS`/`LINES` go with it so Rich reflows instead of clipping.

Content trimming was deliberately left out: it desyncs Rich's cursor state, and `hidden: true` + `!Clear` already covers it.

## Corrections found by checking the page against the code

- **`--share` was undocumented** (`rbx/box/cli.py:669`).
- **`upperValidation` and `lowerViolation` are per-group fields**, not profile-level ones. The page implied otherwise, and hand-transcribed a `groups:` example missing most of what a group record holds — it now defers to the schema.
- **A problem-level `timing.multipliers` cannot switch an environment from a formula to ratios.** The page implied it could; it is an error.
- Dropped the deprecated `timing.multipliers.inferenceTimeout` note — migration trivia that rbx already explains when it sees the old spelling.

## Style guide

Adds §12, *Document the contract, not the implementation*, with a DO/DON'T pair. The `default_timing_formula()` macro was already this rule in practice without being written down. The rule cuts both ways: flags, file names and YAML keys are the contract and still belong on the page in full.

## Verification

- `uv run pytest tests/casts/` — 79 passed.
- `uv run ruff check .` / `ruff format --check .` — clean.
- `uv run mkdocs build` — no new warnings (the ~30 remaining are pre-existing `plans/` macro errors and autorefs).
- All six pages served and checked at `127.0.0.1:8765`; all four casts embed and the partial renders.
- `record-check` reports only the two deliberate `TODO(record)` markers (both need a live MOJ login) and the two pre-existing asciinema.org casts. No missing casts, no orphans.

## Notes

- **Nine inbound links** across six files pointed at anchors on the old single page; all repointed.
- `tests/casts/test_macro.py` pinned the formula macro to `profiling/index.md`; it now points at `computing.md`.
- **No `limits-editor` cast.** The TUI menu labels it *"Edit limits profiles (in development)"* and its profile list is driven by a `highlighted` watcher rather than Enter; recording it kept hanging on exit. The page mentions the editor and flags it unstable instead.
- `mise run record` resolves to the main checkout rather than a worktree; recordings here were made with `uv run python scripts/record.py <name>`.

https://claude.ai/code/session_01SgQMBGvBvwd8ZQrKH7G5Cu


docs/setters/profiling/profiles.md
```diff
@@ -0,0 +1,165 @@
...
+Profiles are plain YAML, so an editor works. Point it at the schema for completion and
+validation:
+
+```yaml title=".limits/local.yml"
+# yaml-language-server: $schema=https://rsalesc.github.io/rbx/schemas/LimitsProfile.json
+```
+
+There is also a profile editor in the {{rbx}} TUI, reachable from `rbx ui`. It is marked *in
+development*, so prefer the file or `rbx time` for anything you intend to keep.
+
```
rsalesc: Remove the whole section.

docs/setters/profiling/profiles.md
```diff
@@ -0,0 +1,165 @@
...
+## Scaling every limit at once
+
```
rsalesc, line 159: Maybe "Scaling every limit locally"

docs/setters/profiling/profiles.md
```diff
@@ -0,0 +1,165 @@
...
+1. The flag exists on `rbx` itself as well as on the subcommand. The two mean the same thing,
+   and the subcommand's wins if you pass both.
+
+`rbx run` fails if the profile does not exist, rather than falling back to something you did not
+ask for.
+
+!!! note
+
+    `rbx irun` takes only the long `--profile`. Its `-p` is already `--print`.
+
```
rsalesc: Please remove

docs/setters/profiling/profiles.md
```diff
@@ -0,0 +1,165 @@
...
+└── .limits/
+    ├── local.yml      # (1)!
+    ├── boca.yml       # (2)!
+    └── polygon.yml
+```
+
+1. The default. `rbx run` uses this one unless told otherwise.
+2. Named for a packager rather than for you — see [Profiles and
+   packaging](#profiles-and-packaging).
+
```
rsalesc, line 26: I think this is buggy and not rendering properly.

docs/setters/profiling/index.md
```diff
@@ -1,649 +1,101 @@
...
+environment configures, and writes the result to a **limits profile** — a file you can keep one
+of per judge you ship to.
 
-Even against a single judge, profiling picks a limit generous enough for the intended solutions
-and tight enough to reject the slow ones, instead of leaving you to guess.
+## Motivational problem
 
-## Quick start
+Everything in this section is the same problem, growing. It asks for the number of pairs of
+values in a list that add up to `K`:
```
rsalesc: This paragraph is weird, what "growing" means here.

docs/setters/profiling/index.md
```diff
@@ -1,649 +1,101 @@
...
-```
+{{ asciinema("time-estimate") }}
 
-!!! warning
-    The BOCA packager **requires** a profile named `boca` to exist. If it doesn't, the packager will fail
-    and ask you to run `rbx time -p boca` first.
+The table at the end is the answer. `cpp` gets `100 ms` from its own measurements, `py` gets
+`200 ms` from its own, and `java` — which nothing solves — gets `×2.0 of cpp`, because the
+environment says an unsolved language should follow C++ rather than fall back to the base
+limit.
```
rsalesc, line 62: Please be clear the user shouldn't expect to make much sense of this before reading the follow-up sections, which will explain every mechanism used to estimate those time limits.

docs/setters/profiling/estimating.md
```diff
@@ -0,0 +1,179 @@
...
+
+From there you can regroup to satisfy the bound, press ++f++ to keep the limits anyway, or
+cancel. Nothing is written until you pick one.
+
+Where there is no picker to reopen — under `--auto`, or in a problem with a single language —
+the violation is reported and recorded in the profile, and the limit is written anyway.
+
+Re-checking after a regroup is cheap. A solution that timed out at some limit also times out at
+any lower one, and a solution that finished is answered by arithmetic, so only the solutions
+whose bound went *up* are run again.
```
rsalesc: Remove paragraph.

docs/setters/profiling/estimating.md
```diff
@@ -0,0 +1,179 @@
...
+## Sharing the report
+
+```bash
+rbx time --share png    # or: --share text
+```
+
+Captures the run report and the limits table and copies it to your clipboard, for pasting into
+wherever the argument about a time limit is happening.
+
+## Flags
```
rsalesc: Maybe link to the CLI reference instead of replicating this here, too easy to diverge.

docs/setters/profiling/computing.md
```diff
@@ -0,0 +1,119 @@
...
+
+[`rbx time`](estimating.md) hands the measured timings to whatever rule your **environment**
+configures, and that rule produces the number. There are two, and they are mutually exclusive:
+**ratios**, which bound the limit from both sides, and a **formula**, which computes it from
+below.
+
+Both live in `env.rbx.yml`, because a time limit rule is a property of the contest you are
+setting, not of one problem in it.
+
+## Time limit ratios
```
rsalesc, line 11: Please show a nice three-way inequality in this section as it makes things way easier to understand, a math formula if it makes it easier (tex).

docs/setters/profiling/language-groups.md
```diff
@@ -0,0 +1,88 @@
+# Language groups
```
rsalesc, line 1: This page and the others has several keyboard shortcuts renderings that are not really rendering properly. Please use the mkdocs syntax for rendering keyboard keys / shortcuts.

docs/setters/profiling/remote.md
```diff
@@ -0,0 +1,56 @@
+# Measuring on the judge itself
```
rsalesc, line 1: {{rbx}} not rendering properly here.

docs/setters/profiling/language-groups.md
rsalesc: Real bug, thanks — `pymdownx.keys` was never enabled, so `++r++` was being emitted as literal text. Enabled it in `mkdocs.yml`; the shortcuts now render as `<kbd class="key-r">` and friends. (The rest of the docs use raw `<kbd>` tags, which is why nothing had caught it before.)

docs/setters/profiling/remote.md
rsalesc: Fixed. Root cause was the include mechanism: `pymdownx.snippets` inlines `--8<--` during Markdown rendering, which happens *after* mkdocs-macros has already run over the page, so any macro inside a shared partial was never expanded. Switched the partial to mkdocs-macros' own `{% include %}` (with `include_dir: docs`), which runs in the same Jinja pass — verified zero literal `{{rbx}}` left and the styled links render on both remote pages.

docs/setters/profiling/profiles.md
rsalesc: Confirmed and fixed — good catch. The fence had no language, and Material only resolves `# (1)!` annotations in a highlighted block, so the markers were printed verbatim. Replaced them with plain inline comments in the tree plus one sentence underneath. (The annotations in the yaml/bash blocks elsewhere in the section are fine — those resolve client-side, same as on the existing pages.)

docs/setters/profiling/profiles.md
rsalesc: Removed the whole "Editing a profile by hand" section, including the TUI limits-editor mention.

docs/setters/profiling/profiles.md
rsalesc: Renamed to "Scaling every limit locally" — better, since the point is that it does not follow you into a package.

docs/setters/profiling/profiles.md
rsalesc: Removed.

docs/setters/profiling/estimating.md
rsalesc: Removed.

docs/setters/profiling/estimating.md
rsalesc: Agreed — replaced the table with a link to `rbx time` in the CLI reference, which is generated from the command so it cannot drift. The flags worth explaining (`--auto`, `--runs`, `--skip-slow`, `--share`, `--runner`) keep their own sections above; the reference carries the exhaustive list. This is the same rule I wrote into the style guide in this PR, so the table was inconsistent with it anyway.

docs/setters/profiling/index.md
rsalesc: Fair — "the same problem, growing" was doing no work. Rewritten to say what it is actually for: every page in the section teaches against one problem so the reader never has to re-orient, and it is solved three ways.

docs/setters/profiling/index.md
rsalesc: Added a note right under the recording saying it is the whole mechanism at once, is not meant to be readable on a first pass, and that the prompt, the bucketing screen, the arithmetic and the final check each get a page below — read it as *what the command does* and come back.

docs/setters/profiling/computing.md
rsalesc: Added. The general form, then the same thing with the numbers from the recording:

```text
acToTimeLimit × (slowest accepted)  ≤  time limit  ≤  (fastest too-slow) ÷ timeLimitToTle
        2.0 × 100 ms                ≤    200 ms    ≤        300 ms ÷ 1.5
              200 ms                ≤    200 ms    ≤        200 ms
```

It lands on 200 ms from both sides, which makes the squeeze obvious.

One thing I did **not** do: render it as real TeX. The site has no math support at all today — no `pymdownx.arithmatex`, no MathJax/KaTeX — so that would mean adding the extension plus a JS bundle site-wide for this one formula. The plain block above matches how the page already shows the default formula. Happy to wire up MathJax properly if you want real math elsewhere too; say the word.

Issue #725, opened 2026-08-24
Research a sturdier way to measure a sandboxed program's peak memory


rsalesc:
Following #720, the memory a sandboxed program is charged on Linux is now a *lower* bound rather than an inflated one, but it is still not a real measurement. It would be good to find something sturdier.

## Where we landed

`ru_maxrss` from `os.wait4()` cannot be trusted on Linux: a forked child's `mm->hiwater_rss` is seeded from the parent's current RSS and `execve` folds that mark into the new process' accounting, so the value is `max(parent_rss_at_fork, child_peak)`. #720 works around this by taking the parent's RSS before the fork as a baseline:

- above the baseline, `ru_maxrss` is the child's peak, exactly;
- at or below it, we fall back to the peak RSS sampled from the child by the alarm thread every 0.3 s.

## What is still weak

- **The fallback is a lower bound.** A program whose peak stays under the parent's footprint is reported as whatever the sampler happened to catch. A spike between two ticks is invisible.
- **It is sampling, at 0.3 s.** Short programs get one or two samples; a program that allocates and frees between ticks reads low.
- **The accuracy depends on the parent.** The fatter the rbx process, the wider the band in which we are guessing. That is backwards — measurement quality should not depend on what the setter's session happens to be holding.
- **`posix_spawn` is not an escape hatch.** It was measured under #720 and is contaminated the same way.

## Directions worth researching

- **A spawner (zygote) process.** Fork a small, near-empty helper once at startup and have it fork/exec the programs and report their `rusage` back. The grandchild's `mm` is copied from the tiny helper, so `ru_maxrss` becomes the child's own peak, exactly, with no baseline arithmetic. Cost: an extra process and an IPC channel, plus care around cwd, fds, pgid, and the coordinated-run path in `stupid_sandbox`.
- **cgroup v2 `memory.peak`.** An exact per-program peak, read after exit. Needs a writable cgroup per run, which is a real constraint on a developer laptop and inside containers; worth checking what is available unprivileged with cgroup v2 delegation, and whether it is worth having only where it works.
- **`/proc/<pid>/status` `VmHWM`.** Exact while the process lives, gone once it is reaped. Could be read at the last possible moment (e.g. from a `PTRACE_O_TRACEEXIT` stop or a `pidfd` + exit-notification dance) rather than sampled.
- **Faster or adaptive sampling** as a cheap improvement to what exists — shorter interval early in a run, or `pidfd`-driven sampling — if the structural options prove too invasive.
[...]

PR #726, opened 2026-08-24
fix(grading): stop charging sandboxed programs the parent's memory

### LINKED ISSUE #720: Sandboxed runs report memory-limit-exceeded on Linux CI (68 of 79 failures)
On Linux CI, anything that actually compiles and runs a program in the sandbox comes back `memory limit exceeded`. This is the single largest cause of CI failures: **68 of the 79** failing tests on `main` trace back to it.

Observed on a dispatched `tests.yml` run against `main` at `f1ee9d52` (run [32678437778](https://github.com/rsalesc/rbx/actions/runs/32678437778)).

## Symptom

A trivial Python hello-world exits cleanly and is still assigned MLE:

```
assert evaluation.log.exitcode == 0          # passes
assert evaluation.log.exitstatus == 'ok'     # fails: 'memory limit exceeded'
```

Because `exitcode` is 0, the verdict is being assigned *after* a clean run — nothing was killed. The check is the post-exit one in `rbx/grading/judge/program.py:329`:

```python
if (
    self.params.memory_limit is not None
    and memory_used > self.params.memory_limit * 1024 * 1024
    or self._alarm_msg == 'memorylimit'
):
```

So one of the two operands is wrong: either `memory_used` is inflated, or `memory_limit` is far below the package's 256 MB.

## Blast radius

The verdict is not confined to the tests that assert on it. `generate_output_for_testcase` treats any non-ACCEPTED outcome as fatal and raises `typer.Exit(1)` (`rbx/box/generators.py:629`), so ~50 unrelated tests across MOJ packaging, statement markdown export, runners, validators and solutions die with a bare `click.exceptions.Exit: 1` that says nothing about memory. Affected files:

- `tests/rbx/box/statements/test_markdown_export.py` (16)
- `tests/rbx/box/packaging/moj/test_statement.py` (13)
- `tests/rbx/box/solutions_test.py` (11)
- `tests/rbx/box/runners/test_local_runner.py` (5)
- `tests/rbx/box/validators_test.py`, `runners/test_capabilities.py`, `packaging/moj/test_packager.py`, `tasks_test.py` (4 each)
- `tests/rbx/box/generator_outputs_test.py` (5), `packaging/moj/test_runner_package.py` (1)

## What has been ruled out

- **Not macOS.** `tests/rbx/box/tasks_test.py` is 9-passed locally on macOS; the darwin branch of `get_memory_usage()` is a different formula, so only Linux is affected.
- **Not the extra rusage terms.** The Linux branch sums `ru_maxrss + ru_ixrss + ru_idrss + ru_isrss`. Measured in a `python:3.14-slim` container, the latter three are all `0` and the total for a hello-world is 12.7 MiB — well under any sane limit. The arithmetic is sound.
- **Not the rusage source.** `os.wait4(self.pid, 0)` is per-child, not `RUSAGE_CHILDREN`, so it is not accumulating other processes.

One suggestive data point: a `g++` invocation that fails instantly on a missing source is reported as `memory: 309MB`, which is implausible for a compiler that never got started. That points at the measurement side rather than the limit side, but it is not conclusive.

## Suggested next step

On a Linux runner, log the raw `ru_maxrss` alongside the effective `params.address_space` for one failing testcase. That single line should settle which operand is wrong and turn this into a one-line fix.

## Why it matters beyond the count

Until this is fixed, CI cannot go green, and it also *masks* other failures in all the files above — so no other fix in those files can be confirmed. #719 clears the other six root causes; this is the one left.

For context: the `tests` workflow has been red since 2025-09-16 and does not run on pushes to `main` (the push trigger is commented out in `.github/workflows/tests.yml`), so failures have accumulated unobserved.

rsalesc:
Fixes #720.

## What was happening

On Linux, a forked child's `mm->hiwater_rss` is seeded from the parent's current RSS, and `execve` folds that high-water mark into the new process' accounting. So the `ru_maxrss` that `os.wait4()` reports for a sandboxed program is really `max(parent_rss_at_fork, child_peak)` — it is not the child's memory.

A pytest worker holding ~300 MB therefore charged **every** program it ran ~300 MB, and everything running under a 256 MB problem limit came back `memory limit exceeded` while exiting cleanly with `exitcode == 0`. That is 68 of the 79 failures on `main`, most of them surfacing as a bare `click.exceptions.Exit: 1` from `generate_output_for_testcase`.

Darwin resets the high-water mark at exec, which is why this never reproduced locally.

Measured with the same script on both platforms, running `true`:

| | parent maxrss | child maxrss |
|---|---|---|
| Linux (`python:3.13-slim`) | 13 MB | 12 MB |
| Linux, after the parent touches 400 MB | 423 MB | **412 MB** |
| macOS | 13 MB | 0.92 MB |
| macOS, after the parent touches 400 MB | 413 MB | 0.92 MB |

This also explains the loose end in the issue: a `g++` that dies instantly on a missing source reporting 309 MB is not a bug in the compiler path — 309 MB *was* the worker's RSS. And the issue's own container check came back clean at 12.7 MiB only because it ran from a bare interpreter; the contamination is invisible unless the parent is fat. Importing `pytest` plus a handful of `rbx.box` modules already high-waters at ~123 MiB before any test does work.

## The fix

Take the parent's RSS immediately before forking as a baseline:

- **Above the baseline**, `ru_maxrss` can only have come from the child, so it is the peak, exactly — unchanged behaviour.
- **At or below it**, the value carries no information about the child, so fall back to the peak RSS sampled from the child itself by the existing alarm thread (which now samples on every tick, and once up front, rather than only when a memory limit is set).

The fallback is a lower bound rather than an exact peak, but it is never inflated by whatever the parent happened to be holding, and any program that actually approaches its limit lands on the exact path. Samples at or above the baseline are discarded, since a sample racing a not-yet-exec'd child would still show the parent's copy-on-write footprint.

Because that sampling is now load-bearing — it is the only measurement of a program whose peak stays under the parent's footprint — it also had to get finer. The first CI run on this branch caught the consequence: at one sample every 0.3 s, `test_run_with_memory_limit_regression_test` (100 MB allocated under a 50 MB limit, held for 0.1 s) slipped through as `ok`. RSS is now sampled every 0.1 s, and CPU time keeps its own 0.3 s cadence, where a limit measured in seconds needs no such resolution.

Sampling cannot be made precise at any interval — a program that allocates and frees between two ticks is invisible either way — so 0.1 s is a compromise rather than a guarantee, and the comment says so instead of implying a bound it cannot honour.

On Darwin nothing changes: `maxrss_inherits_parent()` is false and `ru_maxrss` is used as before.

This also fixes reported memory across the board on Linux, which until now read as roughly the parent's RSS for every solution.

## Tests

`TestMaxrssCorrection` in `tests/rbx/grading/judge/test_program.py` covers both branches of the correction, the pre-exec sample guard, the platform predicate, and the two sides of the behaviour, both run from a parent deliberately holding 150 MB:

- a hello world under a 64 MB limit must **not** come back MLE (the bug);
- a program peaking at 96 MB under a 32 MB limit must **still** be caught (the false negative the first CI run exposed) — that one sits in exactly the band where `ru_maxrss` carries no information about the child, so only the sampler can see it.

`tests/rbx/grading` is 404 passed / 1 skipped on macOS.

On Linux, where the bug actually lives, this branch's CI run takes the suite from **79 failures to 35**, and no test reports `memory limit exceeded` any more. Everything the issue attributed to this root cause is gone: `solutions_test`, `runners/`, `validators_test`, `tasks_test`, `generator_outputs_test`, and `steps_run_test`. The 35 that remain are a different root cause — 34 are `pandoc not found` (`rbx/tooling.py` exits 1; the runner has no pandoc installed, which the MLE cascade was masking) and 1 is a timing assertion in `tests/casts/test_engine.py`. Neither is touched by this PR; they are tracked in #730.

## Follow-up

The fallback is the best estimate available without changing how programs are spawned. #725 tracks researching a sturdier measurement (a small spawner process so the fork is cheap, or cgroup v2 `memory.peak`).

### DIFF
--- a/rbx/grading/judge/program.py
+++ b/rbx/grading/judge/program.py
@@ -16,6 +16,16 @@ from rbx.utils import PathOrStr
 
 FileLike = Union[PathOrStr, IO[bytes], int]
 
+# How often the alarm thread samples the child's RSS. Where ru_maxrss cannot be
+# trusted (see `maxrss_inherits_parent`) this sampling is the only measurement
+# of a program that stays under the parent's footprint, so it has to be fine
+# enough to catch a program that allocates and exits in a few hundred
+# milliseconds. A sample costs a few microseconds.
+RSS_SAMPLE_INTERVAL = 0.02
+# CPU time is read on its own, coarser cadence -- it is compared against a limit
+# in seconds and needs no such resolution.
+CPU_CHECK_INTERVAL = 0.3
+
 
 def _maybe_close_files(files):
     for fobj in files:
@@ -316,16 +326,20 @@ class Program:
             self._sample_rss(process)
         except psutil.Error:
             return
-        while not self._stop_alarm_handler.wait(0.3):
+        elapsed_since_cpu_check = 0.0
+        while not self._stop_alarm_handler.wait(RSS_SAMPLE_INTERVAL):
             try:
                 if process.create_time() != create_time:
                     return
-                if self.params.time_limit is not None:
-                    times = process.cpu_times()
-                    cpu_time = times.user + times.system
-                    if cpu_time > self.params.time_limit:
-                        self._alarm_msg = 'timelimit'
-                        self._kill_process()
+                elapsed_since_cpu_check += RSS_SAMPLE_INTERVAL
+                if elapsed_since_cpu_check >= CPU_CHECK_INTERVAL:
+                    elapsed_since_cpu_check = 0.0
+                    if self.params.time_limit is not None:
+                        times = process.cpu_times()
+                        cpu_time = times.user + times.system
+                        if cpu_time > self.params.time_limit:
+                            self._alarm_msg = 'timelimit'
+                            self._kill_process()
                 memory_used = self._sample_rss(process)
                 if (
                     self.params.memory_limit is not None
--- a/rbx/grading/judge/program.py
+++ b/rbx/grading/judge/program.py
@@ -131,6 +131,65 @@ def get_memory_usage(ru: resource.struct_rusage) -> int:
     return (ru.ru_maxrss + ru.ru_ixrss + ru.ru_idrss + ru.ru_isrss) * 1024
 
 
+def maxrss_inherits_parent() -> bool:
+    """Whether a child's ru_maxrss is contaminated by the parent's memory.
+
+    On Linux, a forked child starts out with its `mm->hiwater_rss` seeded from
+    the parent's current RSS, and `execve` folds that high-water mark into the
+    new process' accounting -- so the `ru_maxrss` reported by `os.wait4()` is
+    really `max(parent_rss_at_fork, child_peak)`. A trivial program forked from
+    a fat parent is charged the parent's whole footprint.
+
+    Darwin resets the high-water mark at exec, so the value is the child's alone.
+    """
+    return sys.platform != 'darwin'
+
+
+def is_own_rss(rss: int, rss_baseline: int) -> bool:
+    """Whether an RSS sample reflects the child rather than the inherited parent.
+
+    A sample taken before the child has exec'd still shows the parent's
+    copy-on-write footprint, so anything at or above the baseline is not
+    trustworthy as the child's own -- those are the cases `ru_maxrss` measures
+    exactly anyway.
+    """
+    return not maxrss_inherits_parent() or rss < rss_baseline
+
+
+def get_child_memory_usage(
+    ru: resource.struct_rusage, rss_baseline: int, sampled_peak: int
+) -> int:
+    """Peak memory of a child, in bytes, with the parent's share removed.
+
+    Where `ru_maxrss` is `max(parent_rss_at_fork, child_peak)` (see
+    `maxrss_inherits_parent`), a value above the baseline can only have come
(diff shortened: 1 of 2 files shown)


rbx/grading/judge/program.py
```diff
@@ -16,6 +16,16 @@
 
 FileLike = Union[PathOrStr, IO[bytes], int]
 
+# How often the alarm thread samples the child's RSS. Where ru_maxrss cannot be
+# trusted (see `maxrss_inherits_parent`) this sampling is the only measurement
+# of a program that stays under the parent's footprint, so it has to be fine
+# enough to catch a program that allocates and exits in a few hundred
+# milliseconds. A sample costs a few microseconds.
+RSS_SAMPLE_INTERVAL = 0.02
```
rsalesc: 0.1 is fine, unfortunately we can't be precise here.

rsalesc: Agreed — dropped to `0.1` in ac02f568. You are right that the finer cadence was buying resolution it cannot honour: a program that allocates and frees between two ticks is invisible at any interval, so 20ms was paying a real cost for a false sense of precision. Reworded the comment to say that plainly and point at #725 rather than claiming the interval is "fine enough to catch" anything.

PR #770, opened 2026-08-24
feat(moj): fill the package author from vars.author

rsalesc:
MOJ hard-requires a non-empty `author` file, and rbx always wrote `Unknown` into it.

This reads the `author` **package var** instead — the same var a statement renders as `\VAR{author}`, so the name in the package and the name in the statement cannot drift apart:

```yaml title="problem.rbx.yml"
vars:
  author: "Ada Lovelace"
```

Any primitive is stringified (`vars` is untyped by design, and a name that parses as a number is still a name). A var that is absent, or blank once stripped, falls back to the `Unknown` placeholder — an empty `author: ""` must not travel through as an empty file that `validate-problem.sh` rejects.

Also documented in `docs/setters/packaging/moj.md` and `rbx/box/packaging/moj/CLAUDE.md`, with tests for the three cases (var present / absent / blank).

Verified: `uv run pytest tests/rbx/box/packaging/moj/` → 171 passed, 4 skipped.

https://claude.ai/code/session_013st32U3G2vyDrtDmAi7q9L

### DIFF
--- a/docs/setters/packaging/moj.md
+++ b/docs/setters/packaging/moj.md
@@ -90,22 +90,6 @@ A few things to keep in mind:
   externalization produces) is rasterized to PNG for you. That needs `pdftoppm`, from poppler; a
   statement with PDF figures and no poppler installed refuses to package, naming the figures.
 
-## Author
-
-Every MOJ package carries an `author` file, and MOJ refuses one that is empty. {{rbx}} fills it
-from the `author` package variable:
-
-```yaml title="problem.rbx.yml"
-vars:
-  author: "Ada Lovelace"
-```
-
-It is the same variable your statement can render as `\VAR{author}`, so the name in the package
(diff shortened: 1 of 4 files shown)


docs/setters/packaging/moj.md
```diff
@@ -90,6 +90,22 @@ A few things to keep in mind:
   externalization produces) is rasterized to PNG for you. That needs `pdftoppm`, from poppler; a
   statement with PDF figures and no poppler installed refuses to package, naming the figures.
 
+## Author
```
rsalesc: Remove this section.

rsalesc: Removed in 500da8b9 — the docs page is back to what it was; the rationale stays in `rbx/box/packaging/moj/CLAUDE.md`.

PR #787, opened 2026-08-25
fix(moj): diagnose a bash too old to run the CLI

rsalesc:
The MOJ CLI is a bash script that needs **bash >= 4** behind a `#!/usr/bin/env bash` shebang, so the shell that runs it is whichever `bash` comes first on `PATH` — and macOS still ships **3.2** at `/bin/bash`. A setter who hits this watches `moj` die before any network call:

```
moj: preciso de bash >= 4 (macOS: brew install bash e rode com ele)
```

rbx made it worse. `whoami` is the first call every MOJ command makes, and it relabelled *any* failure as a missing session — so the headline read "Could not read your MOJ login. Run `moj login` first." for a command that dies in exactly the same way. A loop rather than a fix.

## What changed

**1. `MojUnsupportedShellError` + `_outdated_bash()`.** On the failure path only (after the 429 check, so a full queue is never misread), rbx reads the script's shebang, resolves `env bash` through `PATH`, and asks that shell `echo "${BASH_VERSINFO[0]}"`. Structural rather than matching the CLI's own Portuguese die message — the argument `_HTTP_STATUS_RE` already makes in this module. A healthy setter never pays for it.

**2. `whoami()` narrowed.** It now rewrites a failure as a missing session only when it matches `_NO_SESSION_RE`, and re-raises `MojNotInstalledError` / `MojUnsupportedShellError` untouched. A 500, a proxy, or a CLI too old for the server keeps its own cause instead of being told to log in. `_NO_SESSION_RE` moved up beside the other patterns, since two functions read it now.

**3. `RBX_MOJ_BINARY`.** `moj_command()` reads it per call and `shlex.split`s it, so an interpreter *and* a script both fit:

```bash
export RBX_MOJ_BINARY="$(brew --prefix)/bin/bash $(command -v moj)"
```

An env var rather than an `env.rbx.yml` key on purpose: which `moj` runs is a property of the machine, and committing one setter's absolute paths would break the package for everyone else. Every error message now prints the real command via `_display()`, so the setter can re-run by hand exactly what rbx ran, and a bad override says so by name instead of "install the CLI".

**Docs.** A `The MOJ CLI needs bash 4 or newer` troubleshooting section in `setters/packaging/moj.md` — including the warning that running `bash /path/to/moj` by hand only fixes the times *you* type the command, since rbx inherits your `PATH` and not your habits — plus a fallback subsection for `RBX_MOJ_BINARY`. Pointers added where the CLI is introduced on that page and in `_partials/moj-backend.md`, shared by the `rbx run` / `rbx time` remote pages.

## Verification

Tests were written first and watched fail (8 red for the right reasons), then implemented to green. Ten new cases, including a fake `bash` that answers the version probe and refuses everything else, so the macOS failure is simulated faithfully rather than asserted about.

- `419 passed, 4 skipped` across `tests/rbx/box/runners/moj/`, `tests/rbx/box/packaging/moj/` and `tests/rbx/box/remote_test.py`
- `ruff check` / `ruff format` clean; `mkdocs build` adds no new warnings, and the new anchor resolves from both including pages

Checked end-to-end against the **real** `moj` on macOS, not only stubs:

- under `PATH=~/.local/bin:/usr/bin:/bin`, `whoami()` now raises `MojUnsupportedShellError` naming `/bin/bash`, bash 3 and the `PATH` fix
- with `RBX_MOJ_BINARY` set to Homebrew's bash on that same stripped `PATH`, it clears the guard entirely and gets far enough to fail on `jq` — which `whoami` passed through as `moj: preciso do 'jq'` rather than relabelling it a missing login, the narrowing working on a failure I hadn't anticipated
- a healthy setup is untouched: `moj_command()` is `['moj']` and `whoami()` returns the real login

https://claude.ai/code/session_01GJCA2jG7a9u6ZE7gk8yLsu

### DIFF
--- a/docs/setters/packaging/moj.md
+++ b/docs/setters/packaging/moj.md
@@ -157,6 +157,10 @@ rbx on A package moj -u
 Uploading goes through the [`moj` CLI](https://github.com/cd-moj/moj-cli), so you have to be
 logged in with `moj login` first. {{rbx}} reuses that session and never handles your credentials.
 
+!!! note "On macOS, check your bash first"
+    The CLI needs bash 4 or newer, and macOS ships 3.2. Installing a newer bash isn't enough on
+    its own -- see [The MOJ CLI needs bash 4 or newer](#the-moj-cli-needs-bash-4-or-newer).
+
 The problem is created on the server if it doesn't exist yet, and is named `<org>#<problem>`.
 You tell {{rbx}} which org to use in your `env.rbx.yml`:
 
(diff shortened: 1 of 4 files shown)


docs/setters/packaging/moj.md
```diff
@@ -318,6 +322,56 @@ code needs a judge account, not a different reference.
...
+```
+
+Then check that you got the one you meant:
+
+```bash
+bash --version   # should say 5.x, not 3.2
+moj whoami
+```
+
+!!! warning
```
rsalesc: Remove this warning

rsalesc: Removed in b4b07079. The fallback section below already makes the same point more briefly ("it does nothing for the times you run `moj` yourself. Prefer the `PATH`"), so nothing is lost.

PR #803, opened 2026-08-28
feat(vscode): show expanded var values inline in statements

### LINKED ISSUE #797: Show the expanded variables in the statements

rsalesc:
Closes #797.

In a statement file, each `\VAR{...}` that names a package var now gets a VS Code inlay hint showing its expanded value, so a setter reading a constraints block sees the actual numbers without opening `problem.rbx.yml`:

```latex
\item $1 \le N \le \VAR{N.max}$        100000
\item $1 \le a_i \le \VAR{A.max}$      1000000000
```

Design and plan: `docs/plans/2026-08-28-vscode-statement-var-hints{-design,}.md`.

## How it works

A new read-only `rbx vars [--json]` prints the problem's expanded vars; the extension spawns it once per package root, caches the result, and invalidates on the existing `problem.rbx.yml` watcher. Values therefore stay correct **while you edit** — they do not come from build artifacts, so no build is required and nothing goes stale.

| piece | file |
| --- | --- |
| `rbx vars [--json]` | `rbx/box/cli/commands/vars_cmd.py` |
| reference scanner (pure, `node --test`) | `vscode/src/rbx/statementVars.ts` |
| payload parser (pure) | `vscode/src/rbx/varsPayload.ts` |
| spawn/discovery, extracted from `visualize.ts` | `vscode/src/rbxProcess.ts` |
| per-root cache | `vscode/src/statementVarsIndex.ts` |
| the provider | `vscode/src/statementVarHints.ts` |

## Deliberate limits

- **Problem-root references only.** `\VAR{g.N.max}` gets nothing: it usually sits in a `\BLOCK{for ...}` loop and renders a different value per group, so one badge would have to lie or name a group. `p.`/`problem.`/`contest.` likewise.
- **The raw value, even under a filter.** `\VAR{N.max | sci}` typesets `10^{5}` but badges `100000`. The filter decides typesetting; the badge answers "what number is this?".
- **Nothing is guessed.** Unknown name, non-dotted expression, commented line, escaped `\\VAR` — no badge. An absent badge is never wrong, and a misspelled var shows up as the one reference on the line without one.
- Only manifest-declared statements and tutorials, so contest-level statements and templates get none.

## Two decisions worth a reviewer's attention

**1. The extension spawns rbx, which the testset-view design forbids.** That rule says the extension is a pure reader "because every rbx invocation has side effects on the package cache and could race a run already in flight". `rbx vars` has no side effects: it calls `find_problem_package_or_die` (a YAML load) and `expanded_vars` (pure computation), and nothing on that path reaches `get_problem_cache_dir`. That is enforced by a test, not just asserted — `test_vars_json_creates_no_cache_dir`. `visualize.ts` already records that the rule's stated rationale is outdated anyway: rbx cache dirs take a *shared* session lock, so concurrent processes cannot corrupt a run.

**2. `--json` emits display strings, not JSON numbers.** `{"N.max": "100000"}`. Found in review: values reach the badge through `JSON.parse`, so a bound like `` py`10**18 + 7` `` came back as `1000000000000000000` — a *silently wrong* badge, the one failure mode the design rules out — and `10**21` rendered as `1e+21`. Python has the exact integer; rendering there is lossless.

## Testing

506 extension tests (`npm test`), 49 Python tests across `vars_cmd_test.py`, `lazy_cli_test.py` and `completion/drift_test.py`. Every guard in the scanner is mutation-tested — each one was verified to make a test fail when removed.

**Not visually verified.** No Extension Development Host was available. Worth a human eye on: badge padding (gaps on both sides, neither doubled, not flush against a closing `$`), and that a statement already open at window load gets its hints once discovery finishes without needing an edit.

## Pre-existing issues found, not fixed here

- `resetRbxExecutables` has **no callers**, so changing `rbx.executable` does not invalidate the resolved-binary cache until a window reload.
- `run()`'s timeout sends SIGTERM without awaiting exit, orphaning a process that ignores it.
- `TestingPreset.initialize` omits `min_version`, so every CLI-driven `testing_pkg` test needs a mock — now nine of them.
- `docs/setters/reference/cli.md` is badly stale: a full regeneration is a 1363-line reformat. Only the `rbx vars` entry was hand-inserted here (+16 lines). Worth regenerating deliberately in its own commit.
- `enabled()` is now the fifth verbatim copy of the same config read, and four `register*` bodies are near-identical — a shared helper is worth a follow-up.

https://claude.ai/code/session_01C9XLMd9MNBmMgAhLp6uJ66

### DIFF
--- a/docs/plans/2026-08-28-vscode-statement-var-hints-design.md
+++ b/docs/plans/2026-08-28-vscode-statement-var-hints-design.md
@@ -108,8 +108,15 @@ Python, one new lazy command:
 
 The `ENTRIES` row is not optional bookkeeping -- `tests/rbx/box/lazy_cli_test.py`
 pins the table against the modules, and `--help` renders from the table without
-importing the module. Output is the flat dotted-key map,
-`{"N.max": 100000, "A.max": 1000000000}`, i.e. `Package.expanded_vars` verbatim.
+importing the module. Output is the flat dotted-key map of
+`Package.expanded_vars`, with every value rendered to its display *string*:
+`{"N.max": "100000", "A.max": "1000000000"}`. Strings, not JSON numbers,
+because `JSON.parse` yields IEEE doubles -- a bound of `` py`10**18 + 7` ``
+would reach the extension as `1000000000000000000` and badge a wrong value,
+and `` py`10**21` `` would badge as `1e+21`. The extension only ever displays
+these, so rendering them in Python (where the int is exact) keeps D5's
+"absent, never wrong" promise. The renderer is `str`, which is what Jinja
+itself calls on an unfiltered value, so a bool badges as `True`/`False`.
 
 Extension, following the pure/impure split every `src/rbx/*.ts` module already
 observes:
--- a/docs/plans/2026-08-28-vscode-statement-var-hints-design.md
+++ b/docs/plans/2026-08-28-vscode-statement-var-hints-design.md
@@ -88,8 +88,14 @@ level.
 This is a property to keep, not merely to observe, so it is a test: running
 `rbx vars` in a package with no `.box/` leaves it with no `.box/`.
 
-`rbx visualize` is the extension's existing precedent for spawning at all
-(`visualize.ts:19`), and brings the discovery path this feature reuses.
+`rbx visualize` is the extension's existing precedent for spawning at all, and
+brings the discovery path this feature reuses. Its header also records that the
+rule's stated rationale is out of date: rbx cache directories take a *shared*
+session lock and support any number of concurrent processes, and a cache is
+emptied only under an *exclusive* lock that waits for or refuses live holders.
+So even a command that did touch the cache could not corrupt a run in flight.
+`rbx vars` does not touch it at all, which makes it a narrower exception than
+the one already taken.
 
 ## D3. Components
 
--- a/docs/plans/2026-08-28-vscode-statement-var-hints-design.md
+++ b/docs/plans/2026-08-28-vscode-statement-var-hints-design.md
@@ -145,14 +145,16 @@ globbing `*.rbx.tex` is not.
 The vars cache memoizes per package root, resolves the binary through
 `rbx/executable.ts` (setting, then `PATH`, then a login-shell probe, validated
 by `--version`, cached per root), and is invalidated by the **existing**
-`**/problem.rbx.yml` watcher and its 200 ms debounce. The spawn is therefore
-paid on save, not on keystroke.
+`**/problem.rbx.yml` watcher. That watcher is un-debounced -- the 200 ms
+debounce belongs to the `.rbx` artifact and `testset.yml` watchers, which are
+not the ones that produce vars -- so invalidation is immediate. The spawn is
+paid on save, not on keystroke, because a save is what the watcher observes.
 
 ## D4. Data flow
 
 ```
 problem.rbx.yml saved
-  -> existing watcher, 200 ms debounce
+  -> existing (un-debounced) problem.rbx.yml watcher
   -> vars cache invalidated for that root
   -> onDidChangeInlayHints fires
   -> VS Code re-requests hints for visible statement editors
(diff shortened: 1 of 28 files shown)


rsalesc:
## Follow-up: filters are now honoured

The first cut badged the **raw** value under a filter — `\VAR{N.max | sci}` showed `100000` while the statement typesets `10^{5}`. That was a retreat from porting `scientific_notation` to TypeScript, which would have duplicated a function that is all edge case.

There was a third option, and it is better than either: **let rbx render the expression, and teach the filter what medium it is rendering for.**

```
\VAR{N.max | sci}     10⁵
\VAR{M.max | sci}     2×10⁵
\VAR{MOD | rsci}      10¹⁸ + 7
\VAR{name | upper}    FOO
```

Every filter works, not just those two — the extension reimplements nothing.

Design: `docs/plans/2026-08-29-statement-filter-targets-design.md`.

### The target belongs to the filter

`add_builtin_filters(env, target)`. A filter's **rules** — when `sci` abbreviates, when it pulls out a multiplier, when it declines — are a property of the number. Only the **spelling** is a property of the medium: `2 \times 10^{5}` for a PDF, `2×10⁵` for an inlay hint that cannot typeset maths.

`MARKDOWN` maps to the LaTeX formatter and is deliberately its own value: a Markdown statement puts constraints in `$…$` math, so LaTeX is right there. Naming it separately makes the day that stops being true a one-line change.

**LaTeX output does not move**, and that is proven rather than asserted: the review extracted the parent implementation and diffed it against the new one over **202,188 comparisons** (9,628 values × 7 `zeroes` thresholds × both `rest` modes, comparing return values *and* raised exceptions) — zero divergences. The golden table was mutation-tested too.

### Only a filtered reference costs a spawn

An unfiltered `\VAR{N.max}` still comes from the bulk `rbx vars --json` map, no process at all. A filtered one needs `rbx vars --render`, which reads expressions from stdin and returns them rendered. Results cache per `(root, expression)` **including failures**, so typing `\VAR{N.max | sc` asks once rather than once per keystroke.

Verified end to end against a real package and a real spawn: three expressions in **one** 350 ms spawn; two spellings of the same pipeline collapsing to one render; a declined expression re-asked in **0 ms** with no spawn.

### Two bugs found on the way

**A silent one, in code already on this PR.** `rbx` ends its stdout with a show-cursor escape whenever `FORCE_COLOR` is set in the environment the editor inherited. `JSON.parse` throws on it, so the hints drew **nothing at all** — on some machines and not others, depending on how the editor was launched. Two earlier reviews had called trailing noise a "considered limit" and "theoretical". It was neither:

```
raw ends with: "647\"}\n[?25h"
parseVarsPayload => undefined      ← no badges, silently
```

Escapes are now stripped, with a regression test built from the captured bytes.

**A protocol violation caught in review.** A canonical expression could contain a newline (`\VAR{N.max | sci(9,\n 3)}`), and the CLI's stdin protocol is one expression per line — rbx would have received two fragments and answered neither. Such references now yield no hint.

### Also here

- `run-extension.sh` pins the development host to the checkout's own `rbx`, via `rbx.executable` in a per-worktree profile rather than a `PATH` export — the `code` CLI hands off to a running instance over IPC, so an exported `PATH` reaches the extension host only when no editor is already open. `--no-pin-rbx` opts out. This matters more now that half the feature shells out.
- A shared `make_jinja_env(target, syntax=...)` factory. `syntax` is a second axis rather than a `markdown` boolean, because the combination the feature needs — LaTeX delimiters with TEXT formatting — is unsayable with a boolean named for the medium.

### Not fixed here

`rsci(-100007)` is `-100007` rather than the negation of `10^{5} + 7`: the negative-number recursion does not forward `rest`. Pre-existing, confirmed against the parent commit, and left alone because this change's constraint is that shipped statement output does not move. It deserves its own commit, its own golden row and a changelog line.

### Testing

524 extension tests, 492 Python (1 skipped), typecheck clean, mkdocs builds. Still **not visually verified** — no Extension Development Host was available.

PR #832, opened 2026-08-31
feat(timing): checksum what a time limit was estimated from

### LINKED ISSUE #823: Checksum solutions used in timelimit estimation

rsalesc:
Closes #823.

## Why

`.limits/<profile>.yml` recorded the time limit `rbx time` computed, but nothing about what it computed it *from*. Edit the accepted solution, add a slow one, rewrite a generator — the limit stays put, indistinguishable from one measured five minutes ago.

Profiles now carry an `estimationChecksum`, and the commands that consume a profile recompute it and say when it no longer matches.

## The checksum

One segmented string — a single field to store and compare, but the segment that differs names the bucket that moved:

```
v1.h.9f3a1c22.4b7e0d81.c1a8f930
│  │ solutions  interactor  tests
│  └ level: `l` or `h`
└ format version
```

**Light** covers the solutions carrying an inference role (`lower`/`upper`), each hashed over its transitive source closure — a C++ solution whose hot loop lives in an included header changes its timing without its own bytes moving. The set is *derived* from the package rather than stored, so the estimator and the checker agree without extra state.

**Heavy** adds the interactor and a digest of every built test input. Built inputs rather than testplan declarations: the inputs are the ground truth the declarations were only trying to produce.

The version prefix means an unknown format compares as "cannot tell", never as a mismatch. Same for an unparseable string — the field is hand-editable YAML, and a garbled one should cost a missing warning, never a failed build.

## When the heavy level is refused

Silently downgraded to light when the digests cannot be trusted:

- no manifest, or one from a different manifest version;
- a build that did not check generator determinism (`-v0`) — an unseeded generator would otherwise mismatch on *every* build and turn the warning into noise;
- a build restricted to a subset of groups (`--samples-only`), whose manifest describes a different testset than the estimate ran against.

## Consumers

| Command | Where | Why there |
| --- | --- | --- |
| `rbx package build` | after `builder.verify` | only the build that just ran leaves a manifest describing the tests actually being packaged |
| `rbx run -p <profile>` | `_set_timing_profile` | naming a profile is asking to be judged by its limits |
| `rbx time` | beside the "Current limits" table | informational — the command exists to replace the estimate |

Always a warning, never an error: only the setter can say whether an edited solution was one whose timing mattered. Profiles written by `--strategy inherit` or a hand-set limit carry no checksum, so they are never checked.

## Notable

The test digests already existed — generation computes them for duplicate/non-determinism detection and threw them away. They are now persisted into `build/testset.yml` (`MANIFEST_VERSION` 1 → 2), which makes the manifest **load-bearing for rbx itself for the first time**; it was written purely for external readers. That is the deliberate cost of making the tests segment free at check time. `write_manifest_or_warn` still downgrades every failure to a warning — a missing manifest costs the checksum its heavy level and nothing else.

## Testing

28 new tests in `tests/rbx/box/estimation_checksum_test.py`. Also ran `test_testset_manifest`, `limits_info_test`, `test_limits_*`, `test_timing*`, `generators_test`, `tests/rbx/box/packaging`, `lazy_cli_test`, `test_schema`, `test_run_cli`, `test_run_profile_cli` — all pass. Two failures in `generators_test` (`test_generator_non_existent`, `test_generator_not_compile`) reproduce identically on unmodified `main`.

Design doc: `docs/plans/2026-08-31-estimation-checksum-design.md`. User docs: a new "When an estimate goes stale" section in `docs/setters/profiling/profiles.md`.

https://claude.ai/code/session_01AMeu5XDGXFpx7bD7TMkj4x

### DIFF
--- a/docs/setters/profiling/profiles.md
+++ b/docs/setters/profiling/profiles.md
@@ -133,11 +133,10 @@ rbx package boca
 
 ## When an estimate goes stale
 
-An estimate is only as good as what it was measured against. Rewrite the accepted solution,
-add a slow one, change a generator — and the number in the profile stays where it is, looking
-as authoritative as the day it was measured.
+An estimate is only as good as what it was measured against. Rewrite a solution, add a slow
+one, change a generator, and the number in the profile stays as it was.
 
-So {{rbx}} records what it measured, alongside the number:
+So {{rbx}} records what it measured, next to the number:
 
 ```yaml title=".limits/boca.yml"
 timeLimit: 1400
@@ -147,7 +146,7 @@ estimationChecksum: 'v1.h.9f3a1c22.4b7e0d81.c1a8f930'
 That string is a checksum of the solutions the estimate was derived from — the accepted ones
 that set the limit and the slow ones that validated it — plus, when the tests had been built,
 the interactor and the test inputs themselves. `rbx run -p`, `rbx package` and `rbx time`
-recompute it and tell you when it no longer matches:
+recompute it and warn you when it no longer matches:
 
 ```
 The time limit saved in profile boca is stale: the solutions it was estimated from
@@ -155,14 +154,7 @@ have changed since it was estimated.
 Re-run `rbx time -p boca` to refresh it.
 ```
 
-It is a warning and never an error. Only you can say whether a solution you just edited was
-one whose timing mattered; {{rbx}} only says that the question is now open.
-
-Two things it deliberately stays quiet about. A profile whose limit you set by hand or
-inherited from the package carries no checksum at all — nothing was estimated, so there is
-nothing to go stale. And a package whose tests are not currently built is checked on its
-solutions alone: on a fresh clone, before any build, an edited generator goes unnoticed until
-you build.
+It is a warning, not an error: whether the change mattered to the timing is your call.
 
 ## Inheriting from the package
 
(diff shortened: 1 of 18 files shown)


docs/setters/profiling/profiles.md
```diff
@@ -131,6 +131,39 @@ rbx package boca
...
+```
+The time limit saved in profile boca is stale: the solutions it was estimated from
+have changed since it was estimated.
+Re-run `rbx time -p boca` to refresh it.
+```
+
+It is a warning and never an error. Only you can say whether a solution you just edited was
+one whose timing mattered; {{rbx}} only says that the question is now open.
+
+Two things it deliberately stays quiet about. A profile whose limit you set by hand or
```
rsalesc: remove this

docs/setters/profiling/profiles.md
```diff
@@ -131,6 +131,39 @@ rbx package boca
...
+add a slow one, change a generator — and the number in the profile stays where it is, looking
+as authoritative as the day it was measured.
+
+So {{rbx}} records what it measured, alongside the number:
+
+```yaml title=".limits/boca.yml"
+timeLimit: 1400
+estimationChecksum: 'v1.h.9f3a1c22.4b7e0d81.c1a8f930'
+```
+
```
rsalesc, line 145: humanize this, less exageration

docs/setters/profiling/profiles.md
rsalesc: Removed. The two caveats it listed (hand-set profiles carry no checksum; an unbuilt package is checked on solutions alone) live in the design doc instead.

docs/setters/profiling/profiles.md
rsalesc: Toned down — dropped "looking as authoritative as the day it was measured" and the "Only you can say whether a solution you just edited was one whose timing mattered" line. The section is 5 paragraphs now instead of 8.

PR #856, opened 2026-09-07
feat(packaging): package communication problems for DOMjudge

rsalesc:
Follow-up to #598: teaches the DOMjudge packager to ship interactive (COMMUNICATION) problems.

## How DOMjudge does interaction

`validation: custom interactive` in `problem.yaml` is not a validator flavour but a **role change**. `ImportProblemService` registers the output validator as the problem's `run` executable (`combined_run_compare`) rather than its `compare` one; the judgehost then renames the built `run` to `runjury`, drops its own `run-interactive.sh` in its place, and that `exec`s `runpipe` to cross the submission's stdin/stdout with the validator's. The shipped program is therefore called as `runjury <input> <answer> <feedbackdir>` — team output on stdin, team input on stdout — and runs outside the chroot, unmeasured, while only the submission is under `runguard` and the time limit.

## What ships

- **Modern interactor** (the default): its checker is the no-op builtin and the schema forbids configuring another, so it already *is* the sole judge locally. It maps one-for-one onto DOMjudge's single validator, and no checker is shipped.
- **Legacy interactor**: writes an output a real checker grades — two programs where DOMjudge runs one. `output_validators/` grows a `build`/`run` pair: `build` compiles both and heredocs out a `run` that runs the interactor against the submission (inheriting the pipes untouched), then, only if it exited 42, the checker with the interactor's output on stdin. `run` drops its own copies of the pipe fds before the checker runs so the submission sees EOF then, rather than blocking until the wall clock kills it.
- **`testlib_patch`** gains an `argc == 5` arm — testlib's original `<input> <output> <answer>` order plus the feedback dir — because the 4-arg patched `registerInteraction` folds `tout` into `feedbackdir/teammessage.txt`, and the chain needs it as a real file.
- **rbx now always writes its own `build`, for batch too.** DOMjudge's auto-detection compiles whichever source `readdir` hands back first: harmless for a lone checker, wrong the moment a flattened `.cpp` dependency or a chained interactor puts a second source in the directory.

## Verification

Against a local DOMjudge 9.0 (`mise run domjudge up`, then `domjudge.sh import`), two packages built from the `tests/e2e/testdata/interactive` fixture — one legacy, one converted to a modern interactor. Both imported cleanly and DOMjudge judged all seven shipped jury solutions exactly as rbx expects:

| solution | rbx expects | DOMjudge (legacy) | DOMjudge (modern) |
| --- | --- | --- | --- |
| `main.cpp` | AC | `correct` | `correct` |
| `af_wa.cpp` | WA | `wrong-answer` | `wrong-answer` |
| `af_ac_pe.cpp` | WA | `wrong-answer` | `wrong-answer` |
| `af_ac_re.cpp` | RE | `run-error` | `run-error` |
| `af_tl_after_ans.cpp` | TLE | `timelimit` | `timelimit` |
| `af_ac_too_many_iter.cpp` | INCORRECT | `wrong-answer` | `wrong-answer` |
| `interactive-binary-search_mm_naive_cin.cpp` | INCORRECT | `wrong-answer` | `wrong-answer` |

`af_tl_after_ans` is the interactor-exits-first case, and `af_ac_too_many_iter` on the legacy package is the *checker* rejecting — which is what proves the chain actually reaches it.

Also: 26 unit tests in `tests/rbx/box/packaging/test_domjudge.py` and a new `package-domjudge` e2e scenario on the `interactive` fixture.

https://claude.ai/code/session_01UUA839TXvBpFJDKyfYHGUo


docs/setters/packaging/domjudge.md
```diff
@@ -56,6 +54,30 @@ checkers (`wcmp`, `ncmp`, …) as well as your own.
 
 Checkers must be written in C++ (testlib).
 
+## Interactive problems
```
rsalesc: Remove this section.

rsalesc: Removed — the whole `## Interactive problems` section is gone in 19a96863. The support table in `index.md` still carries the interactive checkmark for DOMjudge, and the mechanics stay documented in `rbx/box/packaging/CLAUDE.md`.

One thing that went with it: the note that an interactor, like a checker, has to be C++ (testlib) — the packager errors out otherwise. Say the word if you want that as a clause on the existing "Checkers must be written in C++ (testlib)." line.

